PR #309, opened 2022-01-12
AC component testing

chrisFrazier77:
added unit tests for the autocomplete component


packages/snap-preact-components/src/components/Organisms/Autocomplete/Autocomplete.test.tsx
```diff
@@ -2,22 +2,34 @@ import 'whatwg-fetch';
 import { h } from 'preact';
 import { v4 as uuidv4 } from 'uuid';
 import { render } from '@testing-library/preact';
+import userEvent from '@testing-library/user-event';
+
+import { ThemeProvider } from '../../../providers';
 
 import { Autocomplete } from '../../Organisms/Autocomplete/Autocomplete';
-import { createAutocompleteController } from '@searchspring/snap-preact';
```
korgon, line 7: Any particular reason to drop the use of `createAutocompleteController`? This test got a lot more complicated without it... You can pass services to the function in order to utilize the MockClient. For example:
```
const controller = createAutocompleteController({ client, controller: config }, {client: mockClient });
```

korgon (changes requested):
Manually setting urlManager state in these tests feels wrong - these should be removed, and instead the component should be interacted with as a user would (click, hover, setting input, etc...) in order to have the urlManager state changed.

korgon:
Need to fix failing tests :|

packages/snap-store-mobx/src/Autocomplete/Stores/FacetStore.test.ts
```diff
@@ -26,7 +26,7 @@ describe('Facet store', () => {
 	});
 
 	it('uses the search store FacetStore', () => {
-		const searchData = mockData.autocompleteMeta();
+		const searchData = mockData.searchMeta();
```
korgon, line 29: Since this is in Autocomplete store, we should probably use autocomplete responses.

packages/snap-store-mobx/src/Autocomplete/Stores/FacetStore.test.ts
```diff
@@ -43,7 +43,7 @@ describe('Facet store', () => {
 	});
 
 	it('adds a preview function to each facet value', () => {
-		const searchData = mockData.autocompleteMeta();
+		const searchData = mockData.searchMeta();
```
korgon, line 46: Since this is in Autocomplete store, we should probably use autocomplete responses. This may mean you need to add a new data mock.

korgon (changes requested):

korgon (approved):

PR #336, opened 2022-02-01
onClick Props

chrisFrazier77:
adding onClick props on organisms to tie into child components onClick functions

### DIFF
--- a/packages/snap-preact-components/src/components/Organisms/Autocomplete/Autocomplete.stories.tsx
+++ b/packages/snap-preact-components/src/components/Organisms/Autocomplete/Autocomplete.stories.tsx
@@ -241,6 +241,24 @@ export default {
 				type: 'object',
 			},
 		},
+		onFacetOptionClick: {
+			description: 'Custom onClick event handler for facet options.',
+			table: {
+				type: {
+					summary: 'function',
+				},
+			},
+			action: 'onClick',
+		},
+		onTermClick: {
+			description: 'Custom onClick event handler for Suggested & Trending Terms',
+			table: {
+				type: {
+					summary: 'function',
+				},
+			},
+			action: 'onClick',
+		},
 		...componentArgs,
 	},
 };
--- a/packages/snap-preact-components/src/components/Organisms/Autocomplete/Autocomplete.stories.tsx
+++ b/packages/snap-preact-components/src/components/Organisms/Autocomplete/Autocomplete.stories.tsx
@@ -248,7 +248,7 @@ export default {
 					summary: 'function',
 				},
 			},
-			action: 'onClick',
+			action: 'onFacetOptionClick',
 		},
 		onTermClick: {
 			description: 'Custom onClick event handler for Suggested & Trending Terms',
@@ -257,7 +257,7 @@ export default {
 					summary: 'function',
 				},
 			},
-			action: 'onClick',
+			action: 'onTermClick',
 		},
 		...componentArgs,
 	},
--- a/packages/snap-preact-components/src/components/Organisms/Autocomplete/Autocomplete.tsx
+++ b/packages/snap-preact-components/src/components/Organisms/Autocomplete/Autocomplete.tsx
@@ -206,15 +206,7 @@ export const Autocomplete = observer((properties: AutocompleteProps): JSX.Elemen
 			rows: 3,
 		},
 	};
-	const displaySettings = useDisplaySettings(breakpoints);
-	if (displaySettings && Object.keys(displaySettings).length) {
-		const theme = deepmerge(props?.theme || {}, displaySettings?.theme || {});
-		props = {
-			...props,
-			...displaySettings,
-			theme,
-		};
-	}
+
 	const {
 		hideTerms,
 		hideFacets,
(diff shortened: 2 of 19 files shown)


packages/snap-preact-components/src/components/Organisms/Autocomplete/Autocomplete.tsx
```diff
@@ -597,4 +617,7 @@ export interface AutocompleteProps extends ComponentProps {
 	breakpoints?: BreakpointsProps;
 	controller: AutocompleteController;
 	width?: string;
+	onClickResult?: (e) => void;
```
korgon, line 620: This should be removed in favor of using the `Result` component prop via `theme`.

korgon (changes requested):
Looking at other components, the naming standard seems to be `onSomeThingClick`.

packages/snap-preact-components/src/components/Organisms/Autocomplete/Autocomplete.tsx
```diff
@@ -286,31 +285,28 @@ export const Autocomplete = observer((properties: AutocompleteProps): JSX.Elemen
...
-											hideCheckbox: true,
-											hideCount: true,
-											onClick: onClickFacetOption,
-										},
-										facetPaletteOptions: {
-											hideLabel: true,
-											columns: 3,
-											onClick: onClickFacetOption,
-										},
+							theme: {
```
korgon, line 288: Should these just live at the root of `components` in `theme` instead of being a nested `theme` of `facet`.

packages/snap-preact-components/src/components/Organisms/Results/Results.tsx
```diff
@@ -89,7 +89,7 @@ export const Results = observer((properties: ResultsProp): JSX.Element => {
 		};
 	}
 
-	const { disableStyles, className, layout, style, controller } = props;
+	const { disableStyles, className, layout, style, onClick, controller } = props;
```
korgon, line 92: Remove `onClick` here?

korgon (changes requested):

korgon (changes requested):
Working good!

One thing I just noticed in testing is that we should probably use the `onTermClick` on the popular(trending) terms as well:
https://github.com/searchspring/snap/blob/122c003ce9ee990f5508b57c43579b6035e6b8c6/packages/snap-preact-components/src/components/Organisms/Autocomplete/Autocomplete.tsx#L452

I also wanted to verify this in the `Recommendation` component, and it is not working as expected... (interestingly this works fine in `Autocomplete` and `Results` when using the same `theme` prop)
```
<Recommendation controller={controller} theme={{
	components: {
		result: {
			onClick: () => { console.log('you clicked on a recs result... hooray!') }
		}
	}
}}/>
```

I did some debugging on this, and it seems the issue might be our use `onMouseDown` in the `Result` component... when I switch those `onMouseDown` to `onClick` it works, I'm not sure exactly why. Perhaps you can investigate?

korgon (changes requested):
Components changed should have their tests updated with the new props. It may be a good idea to also test the `theme` prop passing of top level components (eg. `Results`, `Facets`, `Autocomplete` and `Recommendations`) down to these modified components.

The component stories and README.md files will also need to be updated detailing these changes.

packages/snap-preact-components/src/components/Organisms/Autocomplete/Autocomplete.stories.tsx
```diff
@@ -241,6 +241,24 @@ export default {
...
 			},
 		},
+		onFacetOptionClick: {
+			description: 'Custom onClick event handler for facet options.',
+			table: {
+				type: {
+					summary: 'function',
+				},
+			},
+			action: 'onClick',
```
korgon, line 251: Rename.

packages/snap-preact-components/src/components/Organisms/Autocomplete/Autocomplete.stories.tsx
```diff
@@ -241,6 +241,24 @@ export default {
...
+		},
+		onTermClick: {
+			description: 'Custom onClick event handler for Suggested & Trending Terms',
+			table: {
+				type: {
+					summary: 'function',
+				},
+			},
+			action: 'onClick',
```
korgon, line 260: Rename.

packages/snap-preact-components/src/components/Organisms/Autocomplete/readme.md
```diff
@@ -205,3 +205,34 @@ See `<Results />` component documentation for further details.
...
 <Autocomplete controller={controller} input={'#searchInput'} breakpoints={breakpoints} />
 ```
+
+
+### onFacetOptionClick
+The `onFacetOptionClick` prop contains a custom onClick event handler for facet options. Function is passed the click event. 
+
+
+```typescript
+const CustomOnClickFunc = (e) => {
```
korgon, line 215: `CustomOnClickFunc` -> `customOnClickFunc`

korgon (changes requested):

korgon (approved):

PR #342, opened 2022-02-03
Recommendation Component Breakpoints

chrisFrazier77:
breakpoints prop supports all props available to carousel and recommendation

### DIFF
--- a/packages/snap-preact-components/src/components/Molecules/Carousel/Carousel.tsx
+++ b/packages/snap-preact-components/src/components/Molecules/Carousel/Carousel.tsx
@@ -58,6 +58,9 @@ const CSS = {
 				display: 'flex',
 				justifyContent: 'center',
 				alignItems: 'center',
+				'&.ss__carousel__next-wrapper--hidden, &.ss__carousel__prev-wrapper--hidden': {
+					display: 'none',
+				},
 			},
 			'.ss__carousel__next, .ss__carousel__prev': {
 				padding: '5px',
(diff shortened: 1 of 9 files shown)


packages/snap-preact-components/src/components/Molecules/Carousel/Carousel.tsx
```diff
@@ -13,6 +14,8 @@ import { Swiper, SwiperSlide } from 'swiper/react';
 import { defined } from '../../../utilities';
 import { Theme, useTheme, CacheProvider } from '../../../providers';
 import { ComponentProps } from '../../../types';
+import { useDisplaySettings } from '../../../hooks/useDisplaySettings';
+import { BreakpointsProps } from '../../Organisms/Results';
```
dkonieczek, line 18: Since `BreakpointsProps` is pretty generic and also being used in these two files, can you move the typing from the results component to the types.ts file?

dkonieczek (changes requested):

dkonieczek (approved):

korgon (changes requested):
I did a quick look on `/products.html` and it doesn't appear that all props are working as expected with this change. For example, I cannot get the pagination to show via breakpoint props. Other props don't seem to work or change when the breakpoints adjust (eg: `title`) due to an error that occurs when resizing.

image


I was testing with the following:

```
<Recommendation controller={controller} breakpoints={{
	0: {
		title: 'one',
		loop: false,
		slidesPerView: 1,
		slidesPerGroup: 1,
		spaceBetween: 0,
	},
	640: {
		title: 'two',
		loop: false,
		slidesPerView: 2,
		slidesPerGroup: 2,
		spaceBetween: 10,
	},
	1024: {
		title: 'three',
		loop: false,
		slidesPerView: 3,
		slidesPerGroup: 3,
		spaceBetween: 10,
		hideButtons: true,
		pagination: true
	},
	1200: {
		title: 'four',
		loop: false,
		slidesPerView: 4,
		slidesPerGroup: 4,
		spaceBetween: 10,
		hideButtons: true,
		pagination: true
	},
}} />
```

It may be a good idea to add to our tests to ensure these are working as expected.

korgon (approved):

PR #344, opened 2022-02-04
Controller context

chrisFrazier77:
context added on the controller (controller.context)
allow individual controller configs to have their own context, use root context when no context is explicitly passed to the controller config.


dkonieczek (changes requested):
Tested locally and functionality is good. 
Can you also update the docs in `snap-preact/README.md` with this change?

dkonieczek (approved):

PR #358, opened 2022-02-16
Finder e2e tests

dkonieczek:


packages/snap-preact-demo/tests/cypress/integration/finder/finder.spec.js
```diff
@@ -0,0 +1,307 @@
...
+							const targetSelector = controller.targeters[id].targets.map((target) => target.selector).join(',');
+
+							cy.on('url:changed', (newUrl) => {
+								const expectedUrl = controller.store.selections.filter((selection) => selection.selected)[0].services.urlManager.href;
+								expect(expectedUrl).to.exist;
+								expect(newUrl).to.contain(expectedUrl.substring(1)); //remove first character '?'
+							});
+
+							// click on find button
+							cy.get(targetSelector).find(config.selectors?.finder?.findButton).should('exist').click({ force: true });
```
korgon, line 300: Do you think we should assert that we ended up on the page specified in `config.url` with applied filters?

korgon (changes requested):
Looking good.

korgon (approved):

PR #369, opened 2022-02-24
Autocomplete Trending Search

chrisFrazier77:
autocomplete trending.showResults config prop for showing trending term results initially.

### DIFF
--- a/packages/snap-controller/src/Autocomplete/AutocompleteController.ts
+++ b/packages/snap-controller/src/Autocomplete/AutocompleteController.ts
@@ -159,7 +159,7 @@ export class AutocompleteController extends AbstractController {
 		}
 
 		//auto select first trending term?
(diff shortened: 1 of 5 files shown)


korgon (changes requested):
In addition to the functionality with the controller setting - it would be powerful to be able to manually invoke searches (without the setting) as outlined in the issue:

```
controller.on('focusChange', async ({ controller }, next) => {
	const store = controller.store;

	if (store.trending?.length && !store.terms?.length) {
		store.trending[0].preview();
	}

	await next();
});
```

^ and not just for trending terms, but any initial search should be possible this way. If this proves to be too difficult to implement now, perhaps it could be done in the future.

korgon (changes requested):
The profiles set to `realtime` are no longer updating with your recent changes - specifically when adding  and removing skus. Also, removing doesn't seem to actually change the cookie anymore.

korgon (approved):

PR #384, opened 2022-02-28
Autocomplete Component Breakpoints Bug

chrisFrazier77:
bugfix for autocomplete breakpoints props, & added a regression test.

### DIFF
--- a/packages/snap-preact-components/src/components/Organisms/Autocomplete/Autocomplete.test.tsx
+++ b/packages/snap-preact-components/src/components/Organisms/Autocomplete/Autocomplete.test.tsx
@@ -70,13 +70,13 @@ describe('Autocomplete Component', () => {
 			input: controller.config.selector,
 		};
 
-		const rendered = render(<Autocomplete {...args} />, { container: document.getElementById('target') });
-
 		const input = document.querySelector('.searchspring-ac');
 		(input as HTMLInputElement).focus();
 
 		// to deal with timeoutDelay setTimeout used in focus event
 		await new Promise((r) => setTimeout(r, INPUT_DELAY));
+
+		const rendered = render(<Autocomplete {...args} />, { container: document.getElementById('target') });
 
 		const autocomplete = rendered.container.querySelector('.ss__autocomplete');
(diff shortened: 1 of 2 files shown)


korgon (changes requested):
In testing manually, there are more problems with the `theme` as well. Some odd bits of code here too... for example, the following `theme` variable is never used.
https://github.com/searchspring/snap/blob/v0.21.0/packages/snap-preact-components/src/components/Organisms/Autocomplete/Autocomplete.tsx#L175

In addition, the deepmerge happening here is not working, as the function is only meant to take two objects.

https://github.com/searchspring/snap/blob/v0.21.0/packages/snap-preact-components/src/components/Organisms/Autocomplete/Autocomplete.tsx#L285-L293

You could alternatively use deepmerge `all` function and passing an array of objects.
https://www.npmjs.com/package/deepmerge

However, this seems to work in its place:
```
const displaySettings = useDisplaySettings(breakpoints) || {};
const theme = deepmerge(themeOverride, deepmerge(props?.theme || {}, displaySettings?.theme || {}));
props = {
	...props,
	...displaySettings,
	theme,
};
```

To see the problem, see that the breakpoint theme facet limits are not applied with the current code:
```
export class Autocomplete extends Component {
	render() {
		const controller = this.props.controller;

		const theme = {
			components: {
				facet: {
					limit: 9,
				},
			},
		};

		const breakpoints = {
			0: {
				columns: 1,
				rows: 1,
			},
			320: {
				columns: 2,
				rows: 1,
				hideFacets: true,
				vertical: true,
			},
			768: {
				columns: 3,
				rows: 1,
				layout: 'list',
			},
			1200: {
				width: 'auto',
				columns: 2,
				rows: 2,
				theme: {
					components: {
						facet: {
							limit: 4,
						},
					},
				}
			},
		};

		return <LibraryAutocomplete controller={controller} input={controller.config.selector} breakpoints={breakpoints} theme={theme} />;
	}
}
```

Please make a test that uses the theme within the breakpoints to ensure this also continues to work as expected.

PR #386, opened 2022-03-01
Coverage report

dkonieczek:
By adding the `projects` property in the base jest config, this allows jest to run package tests in parallel and display a single coverage table. Additionally, a new static html report is generated in the root `coverage` directory which we will eventually save in a future store to track over time. In order to still be able to run a single package's tests, the `--projects=''` flag is now part of each package npm `test` script.


korgon (commented):
This seems to be a good solution to generate a nice visual report. Can we get this report to show up in the doc app?

As far as using this for Snap metrics - I'd prefer to have accessible specific data points that we can use to visualize coverage changes over time (via Grafana or similar) for each package. Is it possible to get the coverage details on a per package basis in JSON format? Maybe a custom reporter could be used?
https://jestjs.io/docs/configuration#reporters-arraymodulename--modulename-options

korgon:
It seems that functionality for testing specific parts of packages has been lost - we can get it back?
image

korgon (changes requested):
Working great.
Can you add `bail` to the root level jest config?

korgon (approved):

PR #399, opened 2022-03-09
Component testing

chrisFrazier77:


packages/snap-preact-components/src/components/Molecules/Slideout/Slideout.test.tsx
```diff
@@ -126,6 +126,52 @@ describe('Slideout Component', () => {
...
+		const overlayElement = rendered.container.querySelector('.ss__slideout.ss__slideout--active');
+		const styles = getComputedStyle(overlayElement);
+		expect(styles.right).toBe('0px');
+		expect(styles.left).toBe('');
+	});
+
+	it('can set custom transitionSpeed', () => {
+		const args = {
+			active: true,
+			transitionSpeed: '0.55s' as SlideDirectionType,
```
dkonieczek, line 144: I don't think this typing should be here, although it's probably not causing an issue.

dkonieczek (changes requested):
Small cleanup things.
Also noticed there are 3 skipped tests throughout. Let's remove these if they're not passing.

dkonieczek (approved):

PR #405, opened 2022-03-11
Fixing `snap-preact` Config for Services

korgon:
* fully wiring config services
* add configuration to `deepmerge` to prevent loss of special object prototypes when passed via config (eg: services like UrlTranslator)

### DIFF
--- a/packages/snap-preact/src/create/createAutocompleteController.ts
+++ b/packages/snap-preact/src/create/createAutocompleteController.ts
@@ -14,7 +14,10 @@ import type { SnapControllerServices, SnapAutocompleteControllerConfig } from '.
 configureMobx({ useProxies: 'never' });
 
 export default (config: SnapAutocompleteControllerConfig, services?: SnapControllerServices): AutocompleteController => {
-	const urlManager = services?.urlManager || new UrlManager(new UrlTranslator(config.url), reactLinker).detach();
+	console.log('AC services?', services);
+	const urlManager = (services?.urlManager || new UrlManager(new UrlTranslator(config.url), reactLinker)).detach();
+
+	console.log('URLAMANATERER! - ', urlManager, urlManager.subscribe);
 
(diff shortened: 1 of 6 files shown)


packages/snap-preact/src/create/createAutocompleteController.ts
```diff
@@ -14,7 +14,10 @@ import type { SnapControllerServices, SnapAutocompleteControllerConfig } from '.
 configureMobx({ useProxies: 'never' });
 
 export default (config: SnapAutocompleteControllerConfig, services?: SnapControllerServices): AutocompleteController => {
-	const urlManager = services?.urlManager || new UrlManager(new UrlTranslator(config.url), reactLinker).detach();
+	console.log('AC services?', services);
```
dkonieczek, line 17: 👀

packages/snap-preact/src/create/createAutocompleteController.ts
```diff
@@ -14,7 +14,10 @@ import type { SnapControllerServices, SnapAutocompleteControllerConfig } from '.
...
+	const urlManager = (services?.urlManager || new UrlManager(new UrlTranslator(config.url), reactLinker)).detach();
+
+	console.log('URLAMANATERER! - ', urlManager, urlManager.subscribe);
```
dkonieczek, line 20: 👀

dkonieczek (changes requested):

packages/snap-preact/src/create/createAutocompleteController.ts
korgon: Derp. Trying to go too fast...

packages/snap-preact/src/create/createAutocompleteController.ts
korgon: Derp. Trying to go too fast...

dkonieczek (approved):

PR #422, opened 2022-03-23
Testing: snap-client

chrisFrazier77:
added missing unit tests to the snap-client

### DIFF
--- a/packages/snap-client/src/Client/Client.test.ts
+++ b/packages/snap-client/src/Client/Client.test.ts
@@ -1,6 +1,9 @@
 import 'whatwg-fetch';
 import { Client } from './Client';
 import type { ClientConfig } from '../types';
+import { MockData } from '@searchspring/snap-shared';
+
+const mockData = new MockData();
 
 const wait = (time = undefined) => {
 	return new Promise((resolve) => {
@@ -101,7 +104,7 @@ describe('Snap Client', () => {
 
 			const fetchApiMock = jest
 				.spyOn(global.window, 'fetch')
-				.mockImplementation(() => Promise.resolve({ status: 200, json: () => Promise.resolve({}) } as unknown as Response));
+				.mockImplementation(() => Promise.resolve({ status: 200, json: () => Promise.resolve({}) } as Response));
 
 			const acparams = {
 				search: {
@@ -183,7 +186,7 @@ describe('Snap Client', () => {
 
 			const fetchApiMock = jest
 				.spyOn(global.window, 'fetch')
-				.mockImplementation(() => Promise.resolve({ status: 200, json: () => Promise.resolve({}) } as unknown as Response));
+				.mockImplementation(() => Promise.resolve({ status: 200, json: () => Promise.resolve({}) } as Response));
 
 			const metaprops = { siteId: '88uyt2m' };
 
@@ -216,9 +219,14 @@ describe('Snap Client', () => {
 
 			const searchRequesterSpy = jest.spyOn(searchRequester, 'request' as never);
 
+			//@ts-ignore
+			const metaRequester = client.requesters.meta.requesters.legacy;
+
+			const metaRequesterSpy = jest.spyOn(metaRequester, 'request' as never);
+
 			const fetchApiMock = jest
 				.spyOn(global.window, 'fetch')
-				.mockImplementation(() => Promise.resolve({ status: 200, json: () => Promise.resolve({}) } as unknown as Response));
+				.mockImplementation(() => Promise.resolve({ status: 200, json: () => Promise.resolve({}) } as Response));
 
 			const searchprops = { siteId: '8uyt2m' };
 
@@ -232,6 +240,21 @@ describe('Snap Client', () => {
 
 			expect(searchRequesterSpy).toHaveBeenCalledTimes(1);
 			expect(searchRequesterSpy.mock.calls).toEqual([[searchparams, searchcacheKey]]);
+
+			const metaRequest = {
+				headers: {},
+				method: 'GET',
+				path: '/api/meta/meta.json',
+				query: {
+					siteId: '8uyt2m',
+				},
+			};
+
+			const metaCacheKey = '/api/meta/meta.json{"siteId":"8uyt2m"}';
+
+			expect(metaRequesterSpy).toHaveBeenCalledTimes(1);
+			expect(metaRequesterSpy.mock.calls).toEqual([[metaRequest, metaCacheKey]]);
+
 			expect(fetchApiMock).toHaveBeenCalledTimes(2);
 			fetchApiMock.mockReset();
 		});
@@ -246,7 +269,7 @@ describe('Snap Client', () => {
 
 			const fetchApiMock = jest
 				.spyOn(global.window, 'fetch')
-				.mockImplementation(() => Promise.resolve({ status: 200, json: () => Promise.resolve({}) } as unknown as Response));
+				.mockImplementation(() => Promise.resolve({ status: 200, json: () => Promise.resolve({}) } as Response));
 
 			const trendingprops = { siteId: '8uyt2m' };
 
@@ -274,7 +297,7 @@ describe('Snap Client', () => {
 
 			const fetchApiMock = jest
 				.spyOn(global.window, 'fetch')
-				.mockImplementation(() => Promise.resolve({ status: 200, json: () => Promise.resolve({}) } as unknown as Response));
+				.mockImplementation(() => Promise.resolve({ status: 200, json: () => Promise.resolve([mockData.recommend()]) } as Response));
 
 			const recommendprops = { siteId: '8uyt2m', tag: 'dress' };
 
@@ -296,7 +319,11 @@ describe('Snap Client', () => {
 
 			expect(recommendRequesterSpy).toHaveBeenCalledTimes(1);
 			expect(recommendRequesterSpy.mock.calls).toEqual([[recommendparams, recommendcacheKey]]);
-			expect(fetchApiMock).toHaveBeenCalledTimes(1);
+
+			//wait for batch timeout
+			await wait(300);
+
+			expect(fetchApiMock).toHaveBeenCalledTimes(2);
 			fetchApiMock.mockReset();
 		});
 	});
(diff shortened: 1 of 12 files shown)


packages/snap-client/src/Client/Client.test.ts
```diff
@@ -34,5 +36,256 @@ describe('SNAP Client', () => {
...
+							},
+						},
+						attributes: {},
+					},
+				],
+				profile: {
+					tag: 'similar',
+				},
+			},
+		];
```
korgon, line 255: This might be a good place to utilize our MockData?

packages/snap-client/src/Client/Client.test.ts
```diff
@@ -34,5 +36,256 @@ describe('SNAP Client', () => {
...
+		//@ts-ignore
+		const metaApi = client.requesters.meta;
+		//@ts-ignore
+		api.request = requestMock;
+		//@ts-ignore
+		api.requesters.suggest.request = requestMock;
+		//@ts-ignore
+		api.requesters.legacy.request = requestMock;
+		//@ts-ignore
+		metaApi.requesters.legacy.request = requestMock;
```
korgon, line 116: This seems to defeat the purpose of using the `jest.mockImplementation`...

packages/snap-client/src/Client/Client.test.ts
```diff
@@ -34,5 +36,256 @@ describe('SNAP Client', () => {
...
+			search: {
+				query: {
+					string: 'hello',
+				},
+			},
+		};
+
+		try {
+			client.autocomplete(acparams);
+		} catch (err) {}
```
korgon, line 155: We should probably prevent this from throwing...

packages/snap-client/src/Client/NetworkCache/NetworkCache.test.ts
```diff
@@ -79,21 +119,247 @@ describe('Network Cache', () => {
...
+
+		const controller = new SearchController(searchConfig, {
+			client: new Client(globals, clientConfig),
+			store: new SearchStore(searchConfig, services),
+			urlManager,
+			eventManager: new EventManager(),
+			profiler: new Profiler(),
+			logger: new Logger(),
+			tracker: new Tracker(globals),
+		});
```
korgon, line 174: This is a good test. However, it is more of an integration test than a unit test - if anything it should belong in the Client test.

packages/snap-client/src/Client/NetworkCache/NetworkCache.test.ts
```diff
@@ -79,21 +119,247 @@ describe('Network Cache', () => {
...
+			expect(UpdatedMemoryCache).not.toEqual({});
+
+			expect(mockStorage[CACHE_STORAGE_KEY]).toBeDefined();
+		});
+
+		it('can use the get function', async () => {
+			// @ts-ignore
+			let memoryCache = cache.memoryCache;
+
+			expect(memoryCache).toBeDefined();
```
korgon, line 227: What does the `memoryCache` have to do with this specific test?

packages/snap-client/src/Client/apis/Abstract.test.ts
```diff
@@ -0,0 +1,199 @@
...
+			maxRetry: 2,
+			cache: CustomCacheConfig,
+		};
+
+		let api;
+		// end set up
+
+		expect(() => {
+			api = new API(new ApiConfiguration(config));
+		}).not.toThrow();
```
korgon, line 80: We should probably prevent this from throwing...

korgon (changes requested):
I could probably spend another few hours adding comments on these tests - but I think instead we should look at this together and pair on getting this straightened out.

packages/snap-client/src/Client/NetworkCache/NetworkCache.test.ts
chrisFrazier77: There is quite a bit of boilerplate and set up needed for this test. All of the mock storage set up and imports needed for the search controller. Do we want to move all of that over to the client test? The mock storage bits would be duplicated. Since they are also used in other tests in this file.

korgon: I think unit tests should only be testing the functionality provided by the `NetworkCache` class - the constructor, and the methods.

In these tests you have setup the controller for: `caches search responses and uses them` and `can be disabled via the config` where you are testing more than the `NetworkCache` class. For example, the disable config test could just pass a config to the constructor that disables the cache without the need for the controller and all of its services. Testing that the client config passes the config to the `NetworkCache` is more of a `Client` or integration test.

These are good tests, but in my opinion do not belong here. I'd suggest putting these in an "integration" test (similar to the [UrlManager](https://github.com/searchspring/snap/blob/main/packages/snap-url-manager/src/integration.test.ts)) - where the entire setup can be tested all together. If you do that, the `metaResponse` used in the mocking should probably be moved to a `MockData` entry.

packages/snap-client/src/Client/apis/Abstract.test.ts
```diff
@@ -0,0 +1,265 @@
...
+
+		//handles 200s
+		const fetchfn200 = jest
+			.spyOn(global.window, 'fetch')
+			.mockImplementation(() => Promise.resolve({ status: 200, json: () => Promise.resolve({ found: true }) } as unknown as Response));
+
+		const setCache = jest.fn();
+		const getCache = jest.fn(() => Promise.resolve({ foundInCache: true }));
+
+		api.cache.set = setCache;
```
korgon, line 209: Overriding class methods under tests is not ideal. Can we use and/or spy on the actual cache set/get methods?

packages/snap-client/src/Client/apis/Abstract.test.ts
```diff
@@ -0,0 +1,265 @@
...
+		//can handle 429s and retry correct amount of times.
+		const fetchfn429 = jest
+			.spyOn(global.window, 'fetch')
+			.mockImplementation(() => Promise.resolve({ status: 429, json: () => Promise.resolve({ broken: true }) } as unknown as Response));
+
+		await expect(async () => {
+			await api.request(context);
+		}).rejects.toBe(429);
+
+		expect(fetchfn429).toHaveBeenCalledTimes(config.maxRetry + 1);
```
korgon, line 263: Perhaps the reason for this being +1 has to do with the retryCount starting at 1 instead of 0 (as mentioned earlier)?

packages/snap-client/src/Client/NetworkCache/NetworkCache.test.ts
```diff
@@ -1,17 +1,48 @@
...
-import { EventManager } from '@searchspring/snap-event-manager';
-import { Profiler } from '@searchspring/snap-profiler';
-import { Logger } from '@searchspring/snap-logger';
-import { Tracker } from '@searchspring/snap-tracker';
-import { SearchController } from '../../../../snap-controller/src';
 import type { SearchControllerConfig } from '../../../../snap-controller/src/types';
+import { NetworkCache } from '../NetworkCache/NetworkCache';
+
+const CACHE_STORAGE_KEY = 'ss-networkcache';
+const metaResponse = {
```
korgon, line 8: Prefer to see MockData used here.

packages/snap-client/src/Client/apis/Recommend.test.ts
```diff
@@ -0,0 +1,163 @@
...
+		});
+
+		expect(requestMock).toHaveBeenCalledWith(requestUrl, params);
+
+		requestMock.mockClear();
+	});
+
+	it('can call batchRecommendations', async () => {
+		let api = new RecommendAPI(new ApiConfiguration({}));
+		const response = [
```
korgon, line 102: Prefer to see MockData used here.

korgon: Doesn't appear that we even need the response data though - so maybe it could just be removed?

packages/snap-client/src/Client/apis/Recommend.test.ts
```diff
@@ -0,0 +1,163 @@
...
+			path: '/boost/8uyt2m/recommend',
+			headers: {},
+			query: {
+				limits: [undefined],
+				siteId: '8uyt2m',
+				tags: ['similar'],
+			},
+		};
+
+		const requestUrl = '/boost/8uyt2m/recommend{"tags":["similar"],"limits":[null],"siteId":"8uyt2m"}';
```
korgon, line 145: This looks more like a cacheKey... When we assert it below maybe it is not being used correctly?

packages/snap-client/src/Client/apis/Recommend.test.ts
```diff
@@ -0,0 +1,163 @@
...
+			.mockImplementation(() => Promise.resolve({ status: 200, json: () => Promise.resolve(response) } as unknown as Response));
+
+		await api.batchRecommendations({
+			siteId: '8uyt2m',
+			tags: ['similar'],
+		});
+
+		//add delay for paramBatch.timeout
+		setTimeout(() => {
+			expect(requestMock).toHaveBeenCalledWith(params, requestUrl);
```
korgon, line 158: Doesn't seem like fetch should not be called with these params - this looks more like what the `request` method on the API would expect. Something odd going on here - all of your other requests have this expectation:
```
toHaveBeenCalledWith(requestUrl, params)
```

korgon (changes requested):
Thank you for continued your work on testing this monster package! Almost there - just a few oddities found.

packages/snap-client/src/Client/apis/Abstract.test.ts
```diff
@@ -0,0 +1,265 @@
...
+		ttl: 2222,
+		enabled: false,
+		maxSize: 4, // KB
+		purgeable: false,
+	};
+
+	// set up
+	const fetchfn = jest
+		.spyOn(global.window, 'fetch')
+		.mockImplementation(() => Promise.resolve({ status: 200, json: () => Promise.resolve() } as unknown as Response));
```
korgon, line 61: I wonder if we can somehow use `Partial` here and wherever else `as unknown as Response` is being used...

packages/snap-client/src/Client/apis/Abstract.test.ts
chrisFrazier77: this +1 logic makes sense to me here. You set the max retry count to 3, expect it to make 4 calls. The config value is maxRetry, not maxAttempts. 3 retries + your initial attempt.

PR #423, opened 2022-03-25
feat(recommendations): add results prop to recommendations component,…

dkonieczek:
… allows for slicing results

### DIFF
--- a/packages/snap-preact-components/src/components/Organisms/Recommendation/Recommendation.tsx
+++ b/packages/snap-preact-components/src/components/Organisms/Recommendation/Recommendation.tsx
@@ -12,7 +12,7 @@ import { Carousel, CarouselProps, defaultCarouselBreakpoints, defaultVerticalCar
 import { Result, ResultProps } from '../../Molecules/Result';
 import { defined } from '../../../utilities';
 import { Theme, useTheme, CacheProvider } from '../../../providers';
-import { ComponentProps, BreakpointsProps } from '../../../types';
+import { ComponentProps, BreakpointsProps, Result as ResultType, InlineBannerContent } from '../../../types';
 import { useIntersection } from '../../../hooks';
 import { useDisplaySettings } from '../../../hooks/useDisplaySettings';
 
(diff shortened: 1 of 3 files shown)


packages/snap-preact-components/src/components/Organisms/Recommendation/Recommendation.tsx
```diff
@@ -72,12 +73,12 @@ export const Recommendation = observer((properties: RecommendationProps): JSX.El
 		throw new Error(`<Recommendation> Component requires 'controller' prop with an instance of RecommendationController`);
 	}
 
-	if (children && children.length !== controller.store.results.length) {
+	if (!results && children && children.length !== controller.store.results.length) {
```
korgon, line 76: There is a case where you may want to pass both `children` and `results` - in this case would we still want to ensure that children and results are of the same length?

Perhaps we could just move this below the `resultsToRender` declaration and compare `children.length` to `resultsToRender.length`?

packages/snap-preact-components/src/components/Organisms/Recommendation/readme.md
```diff
@@ -29,6 +29,19 @@ The `loop` prop enables 'infinite' looping through the result set when swiping o
 <Recommendation controller={controller} loop={true} />
 ```
 
+### results
+The `results` prop specifies a reference to the results store array. This will override any results specified in `children` and `controller.store.results`
```
korgon, line 33: Based on line 185 of Recommendation.tsx, it doesn't seem like this overrides results specified in `children`.

When `children` are utilized, the results data from `results` or `controller.store.results` are used for tracking purposes - which is why it is important that the lengths match that of the `children`. This is pretty important and may need to be mentioned here.

korgon (changes requested):
We could have problems with tracking if results do not have properties necessary for tracking. It is unlikely to be a problem, but do you think we should be ensuring that the results passed in are objects inside of `controller.store.results` ?

korgon (approved):
Looks good!

PR #424, opened 2022-03-25
build: add lerna option to create github release

dkonieczek:


korgon (approved):
Looks good. I'm unsure if the token will have sufficient privileges (or if it is even necessary) - but I suppose we will find out soon enough :)

PR #426, opened 2022-03-29
feat(recommendationinstantiator): add cart param to recs context

dkonieczek:


packages/snap-preact/src/Instantiators/RecommendationInstantiator.tsx
```diff
@@ -135,6 +139,23 @@ export class RecommendationInstantiator {
...
+							cartContents = cartFuncContents;
+						}
+					} catch (e) {
+						this.logger.warn(`Error getting cart contents from function`, e);
+					}
+				} else if (Array.isArray(cart)) {
+					cartContents = cart;
+				}
+				if (Array.isArray(cartContents)) {
+					(this.config.services?.tracker || this.tracker).cookies.cart.set(cartContents);
```
korgon, line 156: I think we ought to be setting `this.tracker` (along with other services) to the config services (if provided) during construction.

```
(this.config.services?.tracker || this.tracker).cookies.cart.set(cartContents);
```
Should become:
```
this.tracker.cookies.cart.set(cartContents);
```

It looks like we are doing this for services throughout the instantiator - but perhaps it is a good time to refactor and make this better.

korgon (changes requested):
I forgot to include a new attribute for click tracking: `ss-track-cart-view` - in v3 when an element is clicked with this attribute it simply tells all recommendations components to refresh/reload.

Another thing I failed to add to the task (sorry) - was the whole bubbling code that checks for parents of the clicked target to see if they have attributes of interest. This is an issue when a span was clicked that was inside of a button/a tag that has the attribute.
https://github.com/searchspring/ajax-catalog-v3/blob/master/src/internal_modules/attribute-tracking.js#L4

korgon (changes requested):
Perfect! This should get things up to date.

Last thing... there are likely tests and/or documentation that should be updated with these changes... can you please ensure they are included in the PR?

docs/INTEGRATION.md
```diff
@@ -79,6 +80,7 @@ Similar to the above context variables however these do not apply globally. They
 | Option | Value | Page | Description |
 |---|---|:---:|---|
 | product | current product sku | product detail page | required if product detail pages contain recommendations |
+| cart | array (or function that returns an array) of current cart skus | all | optional method of setting cart contents |
```
korgon, line 83: Looks like we may be missing `shopper` here?

packages/snap-tracker/src/Tracker.ts
```diff
@@ -130,6 +130,7 @@ export class Tracker {
 					`ss-${this.config.id}-cart-view`,
 					`ss-${this.config.id}-intellisuggest`,
 					`ss-${this.config.id}-intellisuggest-signature`,
+					`href`,
```
korgon, line 133: What is it we are doing with the `href` here?

korgon (changes requested):
Couple things in comments...
Also I find it a bit confusing that we have both `cart` and `shopper.cart` as recommendation contexts. I wonder if we should modify this a bit?
https://github.com/searchspring/snap/blob/628694c0b920f97959f202c536698839cc5d757d/docs/INTEGRATION.md?plain=1#L61-L75

dkonieczek:
Turns out the RecommendationInstantiator doesn't even used shopper.cart, it only uses shopper.id so I removed it from the documentation and example code.

korgon (approved):

PR #448, opened 2022-04-07
Finder persist

dkonieczek:


packages/snap-store-mobx/src/Finder/FinderStore.ts
```diff
@@ -40,11 +52,70 @@ export class FinderStore extends AbstractStore {
...
+			this.storage.clear();
+			this.selections = [];
+			this.loaded = false;
+		}
+	};
+
+	loadPersisted(): void {
+		if (this.config.persist?.enabled && !this.loaded) {
+			const date = this.persistedStorage.get('date');
+			const data = this.persistedStorage.get('data');
```
korgon, line 89: Would this data always be equal to the last selectonStore data?
If so, could we possibly do away with `this.data` altogether?

packages/snap-store-mobx/src/Finder/README.md
```diff
@@ -31,6 +31,35 @@ The meta property is an object containing the meta data retrieved from the Searc
...
+A reference to the [StorageStore](https://github.com/searchspring/snap/tree/main/packages/snap-store-mobx/src/Storage)
+
+## `persistedStorage` property
+A reference to the [StorageStore](https://github.com/searchspring/snap/tree/main/packages/snap-store-mobx/src/Storage) used for persisting finder selections (requires `config.persist.enabled` to be `true`)
+
+## `state` property
+An object used for finder state.
+
+### `persisted` property
+Then `state.persisted` property is used to determine if finder selections have been persisted. (requires `config.persist.enabled` to be `true`)
```
korgon, line 51: The `state.persisted` ...

packages/snap-store-mobx/src/Finder/README.md
```diff
@@ -31,6 +31,35 @@ The meta property is an object containing the meta data retrieved from the Searc
 ## `pagination` property
 See [SearchStore](https://github.com/searchspring/snap/tree/main/packages/snap-store-mobx/src/Search) `pagination` property
 
+## `config` property
+
+A reference to the selection config object that was provided to the [FinderController](https://github.com/searchspring/snap/tree/main/packages/snap-controller/src/Finder)
```
korgon, line 36: Is this a "selection config object" or something else/more?

packages/snap-store-mobx/src/Finder/Stores/SelectionStore.ts
```diff
@@ -156,11 +210,14 @@ class Selection extends SelectionBase {
 
 		this.selected = value;
 		this.storage.set('selected', value);
+		this.state.persisted = false;
 
 		if (!value) {
-			this.services.urlManager.remove(`filter.${this.field}`).go();
```
korgon, line 161: I think we might be able to go back to this one liner.

packages/snap-store-mobx/src/Finder/Stores/SelectionStore.ts
```diff
@@ -156,11 +210,14 @@ class Selection extends SelectionBase {
...
+			this.services.urlManager = this.services.urlManager.remove(`filter.${this.field}`);
+			this.services.urlManager.go();
 		} else {
-			this.services.urlManager.set(`filter.${this.field}`, value).go();
```
korgon, line 163: I think we might be able to go back to this one liner.

packages/snap-store-mobx/src/Finder/Stores/SelectionStore.ts
```diff
@@ -229,9 +287,11 @@ class SelectionHierarchy extends SelectionBase {
 			});
 
 		if (!value) {
-			this.services.urlManager.remove(`filter.${this.field}`).go();
```
korgon, line 232: I think we might be able to go back to this one liner.

packages/snap-store-mobx/src/Finder/Stores/SelectionStore.ts
```diff
@@ -229,9 +287,11 @@ class SelectionHierarchy extends SelectionBase {
...
+			this.services.urlManager = this.services.urlManager.remove(`filter.${this.field}`);
+			this.services.urlManager.go();
 		} else {
-			this.services.urlManager.set(`filter.${this.field}`, value).go();
```
korgon, line 234: I think we might be able to go back to this one liner.

korgon (changes requested):
I noticed a few small things... but overall this is great! Looking forward to getting this released soon!

packages/snap-controller/src/Finder/FinderController.ts
```diff
@@ -74,23 +83,47 @@ export class FinderController extends AbstractController {
 		return params;
 	}
 
-	find = (): void => {
+	find = async (): Promise<void> => {
+		await this.store.save(); // save current selections to storage
+
+		try {
```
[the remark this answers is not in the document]
dkonieczek: Was required this way for [this test](https://github.com/searchspring/snap/pull/448/files#diff-868ba30109addc164c053fe2e278885768cb71da15eb53f0b12422ea2b116cc8R378) however I agree that it doesn't make sense here. I removed the try/catch and also removed the throw so it just returns and prevents the redirection upon middleware error

packages/snap-store-mobx/src/Finder/FinderStore.ts
dkonieczek: I think we should leave this because data contains information such as pagination data which you may always be rendering even if selections were persisted.

packages/snap-controller/src/Finder/FinderController.ts
```diff
@@ -87,23 +87,16 @@ export class FinderController extends AbstractController {
...
+			await this.eventManager.fire('beforeFind', {
+				controller: this,
+			});
 		} catch (err) {
-			if (err) {
-				this.log.error(err);
+			if (err?.message == 'cancelled') {
+				this.log.warn(`'beforeFind' middleware cancelled`);
+			} else {
+				this.log.error(`error in 'beforeFind' middleware`);
```
korgon, line 97: We should probably still log the error itself though (since we are catching it)...

```
this.log.error(err);
```

korgon (changes requested):
Need to log the error :)

packages/snap-store-mobx/src/Finder/FinderStore.ts
korgon: Sounds good!

korgon (approved):

PR #449, opened 2022-04-07
siteId override for branch switch

chrisFrazier77:
attempt to grab the siteId out of the script src before using the siteId in the config.

### DIFF
--- a/packages/snap-preact/src/Snap.tsx
+++ b/packages/snap-preact/src/Snap.tsx
@@ -229,7 +229,7 @@ export class Snap {
 				this.logger.setMode(LogMode.DEVELOPMENT);
 				this.logger.warn(`...loading build... '${branchParam}'`);
 
-				//lets try and get the siteId from the current bundle script in case there its not the same as the config.
+				//lets try and get the siteId from the current bundle script in case its not the same as the config.
 				let siteId;
 				try {
 					const scriptsrc = document.getElementById('searchspring-context').getAttribute('src');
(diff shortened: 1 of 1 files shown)


packages/snap-preact/src/Snap.tsx
```diff
@@ -229,9 +229,26 @@ export class Snap {
 				this.logger.setMode(LogMode.DEVELOPMENT);
 				this.logger.warn(`...loading build... '${branchParam}'`);
 
+				//lets try and get the siteId from the current bundle script in case its not the same as the config.
+				let siteId;
+				try {
+					const scriptsrc = document.getElementById('searchspring-context').getAttribute('src');
```
dkonieczek, line 235: Lets use the same query selector & logic that the getContent function uses to retrieve the script: 
https://github.com/searchspring/snap/blob/867740160cd7df24ff126c3d2068a4d5551d352f/packages/snap-toolbox/src/getContext/getContext.ts#L7-L14

packages/snap-preact/src/Snap.tsx
```diff
@@ -229,9 +229,26 @@ export class Snap {
...
+					siteId = scriptsrc
```
dkonieczek, line 236: Let's use this expression to also match the path before the 6 digit siteId `.*\/[a-zA-Z0-9]{6}\/`

```
const path = 'https://snapui.searchspring.io/abc123/bundle.js'.match(/.*\/[a-zA-Z0-9]{6}\//);
if(path) {
    const src = `${path}${branchParam}/bundle.js`;
}
```

dkonieczek (changes requested):

dkonieczek (approved):

PR #456, opened 2022-04-13
fix(facetslider): prevent chrome from crashing when tickSize <= 0

dkonieczek:


packages/snap-preact-components/src/components/Molecules/FacetSlider/FacetSlider.tsx
```diff
@@ -167,6 +166,13 @@ export const FacetSlider = observer((properties: FacetSliderProps): JSX.Element
 		style,
 	} = props;
 
+	let { tickSize } = props;
+
+	if (isNaN(Number(tickSize)) || Number(tickSize) <= 0) {
+		// fallback to default (causes chrome to crash)
+		tickSize = properties.facet?.step * 10 || 20;
+	}
+
```
chrisFrazier77, line 175: This is a nice add! However it isn't converting the actual tickSize prop value to number anywhere. So if you pass it "20" as a string it still breaks things. I think we should change 
tickSize: tickSize,
to 
tickSize: Number(tickSize), 
down where it gets passed into useRanger.

chrisFrazier77 (changes requested):

chrisFrazier77 (approved):

PR #460, opened 2022-04-14
refactor: profiler typing - enable strict mode, update tests

dkonieczek:


packages/snap-profiler/src/Profiler.ts
```diff
@@ -66,7 +66,7 @@ class Profile<T> {
 	}
 
 	public stop(): Profile<T> {
-		if (!this.time.end) {
+		if (!this.time.end && this.time.begin) {
```
korgon, line 69: Nice catch!

korgon (approved):

PR #468, opened 2022-04-19
refactor(urlmanager): enable strict mode, add typing, update tests

dkonieczek:


packages/snap-url-manager/src/UrlManager/UrlManager.ts
```diff
@@ -287,7 +287,7 @@ function removeArrayDuplicates(array: Array<any> | any): Array<any> | any {
 	if (Array.isArray(array) && array.length) {
 		return array.reduce(
 			(accu, item) => {
-				if (!accu.some((keep) => compareObjects(keep, item))) {
+				if (!accu.some((keep: unknown) => compareObjects(keep, item))) {
```
chrisFrazier77, line 290: not sure if its possible, but would be nice to not have unknown here.

PR #472, opened 2022-04-21
refactor(snap-shared): typescript strict mode

chrisFrazier77:
Snap Shared is done and ready I believe, but because it uses the client as a dependency, this pr needs to wait till after the client is typed.


packages/snap-shared/src/MockData/MockData.ts
```diff
@@ -42,7 +42,7 @@ export class MockData {
 
 	constructor(config?: MockDataConfig) {
 		this.initialConfig = config || defaultConfig;
-		this.updateConfig(this.initialConfig);
+		if (config) this.updateConfig(config);
```
korgon, line 45: This seems to change the functionality a bit - it no longer uses the defaultConfig.

korgon (changes requested):

packages/snap-shared/src/MockClient/MockClient.ts
```diff
@@ -14,7 +14,7 @@ import { TrendingResponseModel } from '../../../snap-client/src/Client/apis';
 export class MockClient extends Client {
 	mockData: MockData;
 
-	constructor(global: ClientGlobals, config: ClientConfig = {}) {
+	constructor(global: ClientGlobals, config: ClientConfig) {
```
korgon, line 17: `config` should be an optional construction parameter.

korgon (changes requested):

korgon (approved):

PR #475, opened 2022-04-26
Client types

chrisFrazier77:


dkonieczek (changes requested):
It looks like something is causing list facets with checkboxes to not display correctly on the demostore. Not sure if related to these changes but I checked develop and it's fine there. 

Screen Shot 2022-05-04 at 10 59 02 AM

dkonieczek (approved):

Issue #447, opened 2022-04-07
Fetch Aborting


chrisFrazier77:
Look into adding a fetch abort controller to each requestor in order to prevent multiple search calls from being made at the same time.

PR #484, opened 2022-05-05
fix(searchrequest): allow queries with leading and trailing spaces

dkonieczek:
This is the quick fix that allows for queries with spaces to be pasted and searched for in autocomplete. 

cc #447

### DIFF
--- a/packages/snap-client/src/Client/transforms/searchRequest.test.ts
+++ b/packages/snap-client/src/Client/transforms/searchRequest.test.ts
@@ -79,7 +79,7 @@ describe('search request search transform', () => {
 		expect(params).toEqual({});
 	});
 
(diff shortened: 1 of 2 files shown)


korgon (dismissed):
Nice find!

packages/snap-client/src/Client/transforms/searchRequest.test.ts
```diff
@@ -88,19 +88,7 @@ describe('search request search transform', () => {
 			},
 		});
 
-		expect(params.q).toEqual('a query');
-	});
-
-	it("accepts extended 'string' syntax", () => {
```
korgon, line 94: This was an oddly named and not so useful test... thanks for removing it.

korgon (approved):

PR #493, opened 2022-05-11
batchRecommendations request concatenates the categories parameter

chrisFrazier77:


packages/snap-client/src/Client/apis/Recommend.ts
```diff
@@ -129,6 +129,15 @@ export class RecommendAPI extends API {
 		const deferred = new Deferred();
 
 		paramBatch.request.tags.push(tag);
+
+		if (categories) {
+			if (!paramBatch.request.categories) {
+				paramBatch.request.categories = [categories];
+			} else {
+				paramBatch.request.categories.push(categories);
```
korgon, line 137: If `categories` is an array we should not be pushing an array into an array :)

Example from your test:
```
request categories: [ [ 'shirts' ], [ 'pants' ] ]
```

We should bolster the type checking here (so that it can catch these things) by adding typing for paramBatch.

korgon (changes requested):

korgon (approved):

PR #506, opened 2022-05-24
merchandising segments

chrisFrazier77:
adding the ability to add merchandising segments via the context

### DIFF
--- a/packages/snap-preact/src/Snap.tsx
+++ b/packages/snap-preact/src/Snap.tsx
@@ -219,7 +219,7 @@ export class Snap {
 		let globalContext: ContextVariables = {};
 		try {
 			// get global context
-			globalContext = getContext(['shopper', 'config']);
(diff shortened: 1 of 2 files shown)


dkonieczek (approved):

packages/snap-preact/src/Snap.tsx
```diff
@@ -237,6 +237,10 @@ export class Snap {
 			throw new Error(`Snap: config provided must contain a valid config.client.globals.siteId value`);
 		}
 
+		if (this.context.merchandising) {
+			this.config.client.globals.merchandising = this.context.merchandising;
```
korgon, line 241: Do we want to pass everything off directly into the `globals.merchandising` here? It may be wise to just pick off the `segments` bit so that we are not locked into this `merchandising` context being only request params?

korgon (changes requested):

PR #508, opened 2022-05-24
Finder Endpoint

chrisFrazier77:
moving finder to use its own requestor to allow hitting finder.json endpoint

### DIFF
--- a/packages/snap-client/src/Client/Client.ts
+++ b/packages/snap-client/src/Client/Client.ts
@@ -104,8 +104,8 @@ export class Client {
 			),
 			finder: new HybridAPI(
 				new ApiConfiguration({
-					origin: this.config.search?.api?.origin,
-					cache: this.config.search?.cache,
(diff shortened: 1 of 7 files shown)


packages/snap-client/src/Client/Client.ts
```diff
@@ -96,6 +102,12 @@ export class Client {
 					cache: this.config.search?.cache,
 				})
 			),
+			finder: new HybridAPI(
+				new ApiConfiguration({
+					origin: this.config.search?.api?.origin,
```
korgon, line 107: Should this use it's own `finder` config instead of `search` ?

korgon (changes requested):
Nice work - just a small comment on the requester config.

Additionally, I think there are some small adjustments that should be made to the README.md file for `snap-client`.

korgon (approved):
Good to go!

PR #512, opened 2022-05-25
refactor merchandising segments

chrisFrazier77:


packages/snap-preact/src/Snap.tsx
```diff
@@ -237,8 +237,14 @@ export class Snap {
 			throw new Error(`Snap: config provided must contain a valid config.client.globals.siteId value`);
 		}
 
-		if (this.context.merchandising) {
-			this.config.client.globals.merchandising = this.context.merchandising;
+		if (this.context.merchandising?.segments) {
+			if (this.config.client.globals.merchandising) {
+				this.config.client.globals.merchandising.segments = this.context.merchandising.segments;
```
korgon, line 242: Just thought of this... but do we perhaps want to deepmerge (to allow both config segments and context segments) ?

packages/snap-preact/src/Snap.tsx
```diff
@@ -239,7 +239,10 @@ export class Snap {
 
 		if (this.context.merchandising?.segments) {
 			if (this.config.client.globals.merchandising) {
-				this.config.client.globals.merchandising.segments = this.context.merchandising.segments;
+				this.config.client.globals.merchandising.segments = deepmerge(
+					this.context.merchandising.segments,
+					this.config.client.globals.merchandising.segments
```
korgon, line 244: Should the order of the deepmerge be the other way around?

korgon (approved):
Looking good. Merge at will.

PR #522, opened 2022-06-07
Removing symver ranges 

chrisFrazier77:
removing symver ranges from all package json dependencies for more reliability


korgon (commented):
While you are here - I think we are safe to update to storybook 6.5.7 now. Looks like it might fix a lot of the bugs we have seen.

https://www.npmjs.com/package/@storybook/preact

korgon (approved):

PR #524, opened 2022-06-08
feat(tracker): add error beaconing

dkonieczek:
Screen Shot 2022-06-08 at 12 46 55 PM


packages/snap-tracker/src/Tracker.ts
```diff
@@ -211,6 +212,45 @@ export class Tracker {
...
+					context: {
+						website: {
+							trackingCode: siteId,
+						},
+					},
+				});
+			}
+			const payload = {
+				type: BeaconType.ERROR,
+				category: BeaconCategory.METRICS,
```
korgon, line 232: I think `category` here is meant to provide a way to quickly distinguish various error types. I'm not sure we'd have any types here, but we'd probably want to use something like `searchspring.js` as the category root, and then be able to add to that as needed in the future. Eg: `searchspring.js.type` or `searchspring.js.syntax` or `searchspring.js.runtime` etc...

packages/snap-tracker/src/Tracker.ts
```diff
@@ -211,6 +212,45 @@ export class Tracker {
...
+					href: data?.href || window.location.href,
+					siteId: data?.siteId || context.website.trackingCode,
+					framework: data?.framework,
+					version: data?.version || version,
+					filename: data?.filename,
+					stack: data?.stack,
+					message: data?.message,
+					colno: data?.colno,
+					lineno: data?.lineno,
+					timeStamp: data?.timeStamp,
```
korgon, line 246: I wonder if the timestamp / date here is redundant - we don't send those with other beacons.

packages/snap-tracker/src/Tracker.ts
```diff
@@ -211,6 +212,45 @@ export class Tracker {
...
+					},
+				});
+			}
+			const payload = {
+				type: BeaconType.ERROR,
+				category: BeaconCategory.METRICS,
+				context,
+				event: {
+					type: data?.type || 'error',
+					userAgent: data?.userAgent || navigator.userAgent,
```
korgon, line 236: We should not assume navigator or window are available here - we will be pulling these globals out in the near future.

packages/snap-tracker/src/Tracker.ts
```diff
@@ -211,6 +212,45 @@ export class Tracker {
...
+					href: data?.href || window.location.href,
+					siteId: data?.siteId || context.website.trackingCode,
```
korgon, line 238: Is `siteId` redundant with the data in context?

packages/snap-preact/src/Snap.tsx
```diff
@@ -211,7 +211,50 @@ export class Snap {
 		return this.controllers[config.id];
 	};
 
+	public beaconErrorEvent = (event: ErrorEvent): void => {
```
korgon, line 214: Can we use `handlers` similar to what was don in AC Controller?

Eg: `handlers.error`

packages/snap-tracker/src/Tracker.ts
```diff
@@ -211,6 +212,45 @@ export class Tracker {
...
+					framework: data?.framework,
+					version: data?.version || version,
```
korgon, line 240: Version likely redundant with `meta`.

korgon (commented):
I did a quick initial review to get a look at this - looking good  so far.

We do need to finalize the payload for the beacon, I've left some comments about removing possible redundant data.

I'm currently working on ensuring the beacon will store these new events.

packages/snap-tracker/src/Tracker.ts
dkonieczek: I removed date. Keeping timeStamp as in this context it contains the elapsed time since page load and when the error occurred

packages/snap-tracker/src/Tracker.ts
dkonieczek: Since we need the type to remain the same across any events we send in the future, I'm renamed `type` to be `BeaconType.METRIC` with a value of `metric`
And rename the category to `BeaconCategory.RUNTIME` with a value of `searchspring.js.runtime`

korgon: Do we want type to be 'error' since that is likely the DB we will be creating?

dkonieczek:
https://github.com/searchspring/datawarehouse/pull/7

packages/snap-tracker/src/BeaconEvent.ts
```diff
@@ -34,6 +32,7 @@ export class BeaconEvent {
 			initiator: {
 				lib: 'searchspring/snap',
 				'lib.version': version,
+				framework: config.framework!,
```
korgon, line 35: 'lib.framework' for consistency?

korgon (changes requested):
Looking good!

packages/snap-tracker/README.md
```diff
@@ -209,23 +209,17 @@ const handleErrorTracking = (event: ErrorEvent) :void => {
     const userAgent = navigator.userAgent;
     const href = window.location.href;
     const siteId = this.config.client.globals.siteId;
```
korgon, line 211: This one too?  ;)

packages/snap-tracker/README.md
```diff
@@ -209,23 +209,17 @@ const handleErrorTracking = (event: ErrorEvent) :void => {
...
-    const framework = 'preact';
     const date = Date.now();
 
     const beaconPayload: TrackErrorEvent = {
-        type,
```
korgon: Removed here but not in de-structure above.

korgon (commented):
Small things I missed but noticed here :)

korgon (approved):
Should we just start sending these even without the beacon part setup? Guess it wont hurt anything eh?

PR #533, opened 2022-06-10
Recommendation fixes

chrisFrazier77:


korgon:
Can we add a test to ensure that when order is used the content that resolves with the request is correct?
(essentially testing what this branch resolves)

korgon:
I'm working on adding a test for this so we can get this in the next release.

korgon (approved):
Good additional test to prevent future regression!

PR #534, opened 2022-06-10
ac no results fix

chrisFrazier77:
checking input state to determine visible bool

### DIFF
--- a/packages/snap-preact-components/src/components/Organisms/Autocomplete/Autocomplete.tsx
+++ b/packages/snap-preact-components/src/components/Organisms/Autocomplete/Autocomplete.tsx
@@ -362,7 +362,7 @@ export const Autocomplete = observer((properties: AutocompleteProps): JSX.Elemen
 		}, []);
 	}
 
-	const visible = Boolean(input === state.focusedInput) && (terms.length > 0 || trending?.length > 0);
+	const visible = Boolean(input === state.focusedInput) && (terms.length > 0 || trending?.length > 0 || state.input);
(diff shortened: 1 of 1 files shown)


korgon (commented):
In my testing of this change I am seeing some strange behavior that me may want to consider before releasing.

Mostly, that the query is still being corrected, and the corrected query is used in the search - so the input and what is shown as 'no results for' are different which is a bad experience.

packages/snap-preact-components/src/components/Organisms/Autocomplete/Autocomplete.tsx
```diff
@@ -518,7 +518,7 @@ export const Autocomplete = observer((properties: AutocompleteProps): JSX.Elemen
 											cloneElement(noResultsSlot, { search, pagination, controller })
 										) : (
 											<>
-												<p>No results found for "{search.query.string}".</p>
+												<p>No results found for "{search.originalQuery.string}".</p>
```
korgon, line 521: If I remember right, it is possible that if the query is not corrected `originalQuery` may not exist - we may have to have both in here for those cases:
```
{ search.originalQuery?.string || search.query.string }
```

korgon (changes requested):

packages/snap-preact-components/src/components/Organisms/Autocomplete/Autocomplete.tsx
chrisFrazier77: yep

PR #535, opened 2022-06-14
fix: load branch override script after branch override component renders

dkonieczek:
Moving the branch override script to load after the branch override component has been rendered. Pretty sure the issue here was timing between the two. 

Note: Can't delete properties off the window so I'm setting `window.searchspring = undefined`

### DIFF
--- a/packages/snap-preact/src/Snap.tsx
+++ b/packages/snap-preact/src/Snap.tsx
@@ -338,9 +338,7 @@ export class Snap {
 							elem
 						);
 
-						if (window?.searchspring) {
-							window.searchspring = undefined;
-						}
+						window.searchspring = undefined;
 						document.head.appendChild(branchScript);
 					}
 				);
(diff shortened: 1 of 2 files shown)


packages/snap-preact/src/Snap.tsx
```diff
@@ -338,6 +337,11 @@ export class Snap {
 							/>,
 							elem
 						);
+
+						if (window?.searchspring) {
+							window.searchspring = undefined;
```
korgon, line 342: Since we are in snap-preact and assume a window, can we just do `delete window.searchspring` and eliminate the if block?

dkonieczek: I removed the `if` check. Getting an error when using `delete` so we need to set this to `undefined`

Also, the action had a failing test. Tried fixing it but it looks like the HEAD request in `getBundleDetails` never resolves/rejects within the jest test and therefore never reaches the new location of the script append.

korgon: Yea, that test may need rewritten to actually return some data to allow the rest of the code to execute.

korgon: We should get this out. I'll try and add this test back in with the `snap-preact` typing branch.

korgon (approved):

PR #543, opened 2022-06-21
chore(swiper.js): upgrading swiper to latest version

chrisFrazier77:


korgon:
Problematic without Jest support of ESM.

PR #546, opened 2022-06-22
refactor(domtargeter.ts): adding ability to auto retarget

chrisFrazier77:


packages/snap-toolbox/src/DomTargeter/DomTargeter.ts
```diff
@@ -27,19 +30,63 @@ export class DomTargeter {
...
+			if (target.autoRetarget) {
+				this.autoRetarget = true;
+			}
+		});
+
+		if (this.autoRetarget) {
+			this.targets.forEach((target) => {
+				const checker = () => {
+					//lets not just keep trying forever
+					if (this.timeoutTime < 1000) {
```
korgon, line 44: Should all targets be using the same timeoutTime? If you had like 5 targets, wouldn't they all bump this over 1000 after the first check?

Perhaps each target should have its own timeout timer - wouldn't need to be a property on the target, but could maybe just be a scoped variable?

packages/snap-toolbox/src/DomTargeter/DomTargeter.ts
```diff
@@ -27,19 +30,63 @@ export class DomTargeter {
...
+			} else {
+				// attempt retarget on DOMContentLoaded
+				this.document.addEventListener('DOMContentLoaded', () => {
+					this.doTheThing();
+				});
+			}
 		}
 	}
 
+	doTheThing(): void {
```
korgon, line 84: I suggest we use a more descriptive name for this function.

packages/snap-toolbox/src/DomTargeter/DomTargeter.ts
```diff
@@ -27,19 +30,63 @@ export class DomTargeter {
...
+						//did we find any targets?
+						if (elems && elems.length) {
+							//has this target been targeted already?
+							const foundOne = elems.filter((elem) => {
+								if (!targetedElems.find((e) => e == elem)) {
+									return true;
+								}
+							});
+							//got anything worth doing the thing for?
+							if (foundOne) {
```
korgon, line 57: Always true?

image

korgon (changes requested):
See comments.

packages/snap-toolbox/src/DomTargeter/DomTargeter.ts
korgon: Isn't `foundOne` always truthy still?

packages/snap-toolbox/src/DomTargeter/DomTargeter.ts
```diff
@@ -27,17 +28,50 @@ export class DomTargeter {
...
+								this.retarget();
+								target.hideTarget && this.unhideTarget(target.selector);
+							} else {
+								//try again soon
+								setTimeout(checker, timeoutTime);
+							}
+						} else {
+							//try again soon
+							setTimeout(checker, timeoutTime);
```
korgon, line 58: This seems redundant, could we combine logic into a single if statement?

packages/snap-toolbox/src/DomTargeter/DomTargeter.ts
```diff
@@ -27,17 +28,50 @@ export class DomTargeter {
...
-				// unhide targets after re-target attempt in DOMContentLoaded
-				this.targets.forEach((target) => target.hideTarget && this.unhideTarget(target.selector));
-			});
-		}
+		this.targets.forEach((target) => {
+			if (target.autoRetarget) {
+				let timeoutTime = 100;
+				const checker = () => {
+					//lets not just keep trying forever - this waits roughly 12 seconds before giving up.
+					if (timeoutTime < 2000) {
```
korgon, line 36: Do you think we should unhide the target after it gives up looking?

```
target.hideTarget && this.unhideTarget(target.selector);
```

korgon (dismissed):
Approved!

korgon (approved):
Thanks!

PR #564, opened 2022-07-06
fix(searchcontroller): remove search.redirectResponse and personaliza…

dkonieczek:
…tion when comparing params


korgon (approved):
Stringy the params!

PR #571, opened 2022-07-08
feat: transform merchandising.triggerCampaigns

chrisFrazier77:
transform merchandising.triggerCampaigns - merchandising.campaigns & add to SearchMerchandisingStore

### DIFF
--- a/packages/snap-client/src/Client/transforms/searchResponse.ts
+++ b/packages/snap-client/src/Client/transforms/searchResponse.ts
@@ -386,13 +386,17 @@ transformSearchResponse.sorting = (response: searchResponseType) => {
 
 transformSearchResponse.merchandising = (response: searchResponseType) => {
 	const merchandising = response?.merchandising;
-
+	const transformedMerchandising: any = merchandising;
 	if (merchandising.content && Array.isArray(merchandising.content) && !merchandising.content.length) {
-		merchandising.content = {};
+		transformedMerchandising.content = {};
 	}
 
+	if (merchandising.triggeredCampaigns) {
+		transformedMerchandising.campaigns = merchandising.triggeredCampaigns;
+	}
+	delete transformedMerchandising.triggeredCampaigns;
 	return {
-		merchandising,
+		merchandising: transformedMerchandising,
 	};
 };
 
--- a/packages/snap-shared/src/MockData/search/ga9kq2/merchandising_page2.json
+++ b/packages/snap-shared/src/MockData/search/ga9kq2/merchandising_page2.json
@@ -1066,6 +1066,7 @@
     }
   ],
   "merchandising": {
+    "campaigns": [],
     "redirect": "",
     "content": {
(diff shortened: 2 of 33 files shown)


packages/snap-client/src/Client/transforms/searchResponse.ts
```diff
@@ -386,13 +386,17 @@ transformSearchResponse.sorting = (response: searchResponseType) => {
 
 transformSearchResponse.merchandising = (response: searchResponseType) => {
 	const merchandising = response?.merchandising;
-
+	const transformedMerchandising: any = merchandising;
```
korgon, line 389: Should be typing to use for this.

korgon: I think we should also refactor this entire block so that only what we define are included as properties on merchandising - not passing through anything else.

Eg:
```
transformSearchResponse.merchandising = (response: searchResponseType) => {
	const merchandising = response?.merchandising;

	const transformedMerchandising = {
		redirect: merchandising?.redirect || '',
		content: merchandising.content?.length ? merchandising.content : {},
		campaigns: merchandising?.triggeredCampaigns || [],
	};

	return {
		merchandising: transformedMerchandising,
	};
};
```

packages/snap-shared/src/MockData/search/ga9kq2/merchandising_page2.json
```diff
@@ -1066,6 +1066,7 @@
     }
   ],
   "merchandising": {
+    "campaigns": [],
```
korgon, line 1069: Should have same response as page1 here I'd think.

packages/snap-store-mobx/src/Search/Stores/SearchMerchandisingStore.ts
```diff
@@ -16,6 +16,11 @@ export type BannerContent = Partial<Record<ContentType, Content>>;
 export class SearchMerchandisingStore {
 	public redirect = '';
 	public content: BannerContent = {};
+	public campaigns: {
```
korgon, line 19: Is there a type to be used from `snapi-types` for this?

packages/snap-store-mobx/src/Search/Stores/SearchMerchandisingStore.ts
```diff
@@ -28,6 +33,9 @@ export class SearchMerchandisingStore {
 					}
 				});
 			}
+			if (merchData.campaigns) {
+				this.campaigns = merchData.campaigns;
```
korgon, line 37: Should we be using a Class for these objects (similar to how content is done) ?

korgon (changes requested):
Nice work. Couple of review comments to look at.

korgon (dismissed):
Approved with inconsequential comment :)

dkonieczek (approved):

PR #577, opened 2022-07-13
v0.31.0 Release

### LINKED ISSUE #539: Swiper Enhancements
Components using Swiper need a way to be able to use additional modules. Ideally this would be a prop or some other method to allow us to not include all possible modules in our components.

### LINKED ISSUE #551: Improve Autocomplete Tests
We are having intermittent failures on the AC tests (controller and component) - we should consider trying to use `waitFor` from '@testing-library/preact' instead of the arbitrary setTimeout.

### LINKED ISSUE #553: Add `merchandising.campaigns` to API and Stores
The current API returns `merchandising.triggeredCampaigns` now - this needs to be added to the Snap API and the Merchandising store.

### LINKED ISSUE #567: Infinite Scroll Restoration
The recent changes we made to the infinite scroll position still hasn't resolved the issue in certain browsers (firefox)

This task is to always scroll to the saved position and set `history.scrollRestoration = 'manual'` in the SearchController (or Snap since it's on the window)

This should be the default functionality when infinite is enabled and add we'll need a new search config setting to opt out of this (ie. disableScrollRestoration)

dkonieczek:
:snap: 0.31.0 (2022-07-13)
:sparkles: Features
- fixed scroll restoration when using SearchController `config.settings.infinite` and can be disabled by setting `settings.infinite.restorePosition = false` - closes #567
- new `modules` prop on Carousel and Recommendation components. Allows for specifying [Swiper Modules](https://swiperjs.com/swiper-api#modules) - closes #539
- new `campaigns` property added to `store.merchandising` object (to indicate if any merchandising campaigns were applied to the search) - closes #553

Other
- improve Autocomplete test stability - closes #551
- rename beacon error `event.timeStamp` to `event.errortimestamp` - closes #572


korgon:
Should add the issues this closes?

PR #579, opened 2022-07-14
Dom Targeter Re-Target

chrisFrazier77:
refactor targetedElems logic to allow adding two elements to the same initial selector

### DIFF
--- a/packages/snap-toolbox/src/DomTargeter/DomTargeter.ts
+++ b/packages/snap-toolbox/src/DomTargeter/DomTargeter.ts
@@ -84,7 +84,9 @@ export class DomTargeter {
 				}
 			});
 
-			targetedElems = targetedElems.concat(elems);
+			if (!target.inject?.element || target.inject.action == 'replace') {
+				targetedElems = targetedElems.concat(elems);
+			}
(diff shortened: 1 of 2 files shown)


packages/snap-toolbox/src/DomTargeter/DomTargeter.ts
```diff
@@ -95,6 +97,11 @@ export class DomTargeter {
 			if (target.inject) {
 				try {
 					const injectedElem = this.inject(elem, target);
+
+					if (target.inject.action !== 'replace') {
+						targetedElems = targetedElems.concat(elem);
```
korgon, line 102: Don't we want to add the `injectedElem` to the `targetedElems` array?

korgon (changes requested):

packages/snap-toolbox/src/DomTargeter/DomTargeter.ts
chrisFrazier77: yeeeep

korgon (approved):

korgon (approved):

PR #580, opened 2022-07-15
UrlManager Docs

chrisFrazier77:
updating url manager docs to filter.fieldName rather than facet.fieldName


korgon (approved):
It is accurate now!

PR #581, opened 2022-07-15
Create Autocomplete Controller Bind

chrisFrazier77:
bind autocomplete controllers even if there are no targets


korgon (approved):
Looks good!

PR #586, opened 2022-07-20
ci: add auto merge main into develop upon publish

dkonieczek:


korgon (approved):
I like this much better than what I had in mind :)

PR #589, opened 2022-07-20
feat: send beacon error events from middleware errors

dkonieczek:


packages/snap-controller/src/Search/SearchController.test.ts
```diff
@@ -597,6 +604,97 @@ describe('Search Controller', () => {
...
+		const handleError = jest.spyOn(controller, 'handleError');
+		const errorLogger = jest.spyOn(controller.log, 'error');
+
+		await controller.search();
+
+		expect(handleError).toHaveBeenCalledWith(error);
+		expect(errorLogger).toHaveBeenCalledWith(error);
+		handleError.mockClear();
+		errorLogger.mockClear();
 	});
```
korgon, line 698: Maybe these tests belong in the Abstract controller tests?

korgon (changes requested):
Looks like a good start for logging these internal errors.

For the generic `handleError` testing, I think putting those tests in the AbstractController test makes most sense (see comment).

korgon (approved):

PR #600, opened 2022-07-28
auto send empty search query if landing page

chrisFrazier77:


korgon:
We will not be adding workaround for this API issue.

PR #625, opened 2022-08-16
Recommendation loop bugfix

chrisFrazier77:


packages/snap-preact-components/src/components/Organisms/Recommendation/Recommendation.test.tsx
```diff
@@ -1,14 +1,423 @@
...
+				profile: {
+					placement: controller.store.profile.placement,
+					tag: controller.store.profile.tag,
+					templateId: 'aefcf718-8514-44c3-bff6-80c15dbc42fc',
+					threshold: 4,
+				},
+			},
+		});
+
+		// // first 4 results should have done impression tracking
```
korgon, line 223: Not a huge deal, but comment doesn't align with slice below.

korgon (commented):
Should we look into restoring the window.IntersectionObserver?

I suppose if it was just straight undefined we wouldn't need to bother. Just don't want it to interfere with any potential future tests using this hook.

korgon (approved):
Good to have testing for this now!

PR #634, opened 2022-08-25
add landingPage to the searchstore

chrisFrazier77:


korgon (changes requested):
We should probably clear `store.landingPage` when a new search does not have a campaign with type of `landing-page`.

korgon (commented):
I'm starting to think we should put this in the merchandising store...

Eg:
```
store.merchandising.landingPage
```

korgon (changes requested):
This is more fitting - thanks for making that change.
We should add the component changes to display this `landingPage.title` when it exists.

korgon (approved):

PR #636, opened 2022-08-30
SiteId Context

chrisFrazier77:
grab and return the siteId from the script context or src

### DIFF
--- a/packages/snap-preact/src/Snap.test.tsx
+++ b/packages/snap-preact/src/Snap.test.tsx
@@ -3,7 +3,7 @@ import 'whatwg-fetch';
 import { cleanup, waitFor } from '@testing-library/preact';
 
 import { MockClient } from '@searchspring/snap-shared';
-import { Tracker } from '@searchspring/snap-tracker';
+import { Tracker, TrackerGlobals } from '@searchspring/snap-tracker';
 import { Logger } from '@searchspring/snap-logger';
 import { cookies } from '@searchspring/snap-toolbox';
 
@@ -177,7 +177,7 @@ describe('Snap Preact', () => {
 			},
 		};
 
-		const tracker = new Tracker(contextConfig.client!.globals);
+		const tracker = new Tracker(contextConfig.client!.globals as TrackerGlobals);
 		const spy = jest.spyOn(tracker.track.shopper, 'login');
 		const snap = new Snap(contextConfig, { tracker });
 		expect(spy).toHaveBeenCalledWith({ id: contextConfig.context.shopper.id });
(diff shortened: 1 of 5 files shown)


packages/snap-toolbox/src/getContext/getContext.ts
```diff
@@ -59,5 +66,14 @@ export function getContext(evaluate: string[] = [], script?: HTMLScriptElement |
 		if (typeof variables[key] === 'undefined') delete variables[key];
 	});
 
+	// if we didnt find a siteId in the context, lets grab the id from the src url.
+	if (!variables[siteIdString]) {
+		const siteIdFromSrc = script.getAttribute('src')?.match(/\/[a-zA-Z0-9]{6}\//);
```
korgon, line 71: I think we may want to make this match query more specific to our script src.

packages/snap-toolbox/src/getContext/getContext.ts
```diff
@@ -33,8 +33,15 @@ export function getContext(evaluate: string[] = [], script?: HTMLScriptElement |
...
 
+	let siteIdString = 'siteId';
+
 	const variables: ContextVariables = {};
 
+	//add siteId to list if its not already in there
+	if (evaluate.indexOf(siteIdString) == -1) {
+		evaluate.push(siteIdString);
+	}
+
```
korgon, line 44: This should probably be opt-in only - and modified in snap-preact to request it.

korgon (commented):
`getContext` should not attempt to get the siteId unless it is requested. This follows similar logic for defaults, like shopper, config and merchandising - it requests the context that it wants.

So we would slightly change the functionality here to look for the siteId only when requested in the function.
```
// get global context
globalContext = getContext(['shopper', 'siteId', 'config', 'merchandising']);
```

packages/snap-toolbox/src/getContext/getContext.ts
korgon: I was just thinking about this a bit more... how do we get the siteId when doing local development? The script src is typically `https://localhost:3333/bundle.js` in these cases... I guess they could add the siteId to the script context?

korgon (commented):
This is it!

Well, a few things that might be worth changing in my review :)

packages/snap-preact/src/Snap.test.tsx
```diff
@@ -383,7 +383,7 @@ describe('Snap Preact', () => {
 				},
 			};
 
-			const client = new MockClient(searchConfig.client!.globals);
+			const client = new MockClient(searchConfig.client!.globals as TrackerGlobals);
```
korgon, line 386: For all of the MockClient changes here - can we switch to typing as `ClientGlobals` instead of `TrackerGlobals`?  They are similar but slightly different.

korgon (changes requested):

korgon (approved):
Looks good!

PR #666, opened 2022-10-11
test(mockdata): refactor tests and stories to use snap-shared Mockdata

chrisFrazier77:


dkonieczek (changes requested):
Changes look good, lets clean up some comments

dkonieczek (approved):

PR #667, opened 2022-10-19
Slideout button

chrisFrazier77:


packages/snap-preact-components/src/components/Molecules/Slideout/Slideout.tsx
```diff
@@ -108,6 +105,29 @@ export function Slideout(properties: SlideoutProps): JSX.Element {
 	);
 }
 
+const ButtonContent = (props: { content: string | JSX.Element | undefined; toggleActive: () => void }): JSX.Element => {
+	const { content, toggleActive } = props;
+
+	if (content && typeof content == 'string') {
+		return (
+			<div
```
dkonieczek, line 113: I think we should keep the `ss__slideout__button` class here to keep backwards functionality. 
Also since it's always a string here, do we need to use `dangerouslySetInnerHTML` instead of just `{content}` in the jsx div? If someone wanted to render html from a string as the button, they could pass html via the the buttonContent prop now. (and use `dangerouslySetInnerHTML` if they had to but it would be at their component)

packages/snap-preact-components/src/components/Molecules/Slideout/Slideout.tsx
```diff
@@ -108,6 +105,29 @@ export function Slideout(properties: SlideoutProps): JSX.Element {
...
+		let elem = () => {
+			let elemm: JSX.Element = cloneElement(content, {
+				onClick: () => toggleActive(),
+			});
+			return elemm;
+		};
+		return elem();
+	} else return <></>;
+};
+
```
dkonieczek, line 130: Haven't tested this but it's interesting that the prop type for `buttonContent` already included `JSX.Element`. Wonder if this was just never working 🤔

packages/snap-preact-components/src/components/Molecules/Slideout/Slideout.tsx
```diff
@@ -1,5 +1,5 @@
 /** @jsx jsx */
-import { h, Fragment, ComponentChildren } from 'preact';
+import { h, Fragment, ComponentChildren, cloneElement } from 'preact';
```
dkonieczek, line 2: Instead of using `cloneElement`, let's use `cloneWithProps` which is already being imported here.

packages/snap-preact-components/src/components/Atoms/Image/Image.stories.tsx
```diff
@@ -4,11 +4,152 @@ import { ArgsTable, PRIMARY_STORY } from '@storybook/addon-docs/blocks';
 
 import { Image, FALLBACK_IMAGE_URL, ImageProps } from './Image';
 import { componentArgs } from '../../../utilities';
-import { MockData } from '@searchspring/snap-shared';
-import { SearchResponseModel } from '@searchspring/snapi-types';
 
-const mockData = new MockData();
-let searchResponse: SearchResponseModel = mockData.search();
+const searchResponse = {
```
dkonieczek, line 8: Good catch here. Can we also include a comment here (and in FacetSlider.stories.tsx) to explain why we needed to hard-code the mock data? (ie. storybook unable to bundle `MockData` due to it using `fs`)

dkonieczek (changes requested):

packages/snap-preact-components/src/components/Molecules/Slideout/Slideout.tsx
chrisFrazier77: Maybe I misunderstood, but I thought the task was to allow the button to be anything, and not force them into the div with the ss-slideout-button class. 

From the ticket - "This is okay if we're adding button styles, but if we want to inherit button styles from a site, it makes it a little more complicated as the trigger is on this specific element." .... "I'd like the ability to change the ss__slideout__button element (from a div to something else) and also add a class or other attributes to this button."

Yeah I fought the second question for a while, but for some reason, rendering {content} inside of a jsx div just was not working when the content was converted from a string originally. It would console log out and look like jsx but it wouldnt render anything. the `dangerouslySetInnerHTML` was the work around I found to work. Happy to take another look if you want to pair on it.

packages/snap-preact-components/src/components/Molecules/Slideout/Slideout.tsx
chrisFrazier77: It was working before, but it would always add the element inside of the wrapping ss__slideout__button div. 

You also could only pass inner text as a string and not html as a string, which wasn't totally necessary but I thought it made it easier to use if it just worked with anything.

dkonieczek (dismissed):

packages/snap-preact-components/src/components/Molecules/Slideout/Slideout.tsx
```diff
@@ -108,6 +105,29 @@ export function Slideout(properties: SlideoutProps): JSX.Element {
...
+					__html: content,
+				}}
+			/>
+		);
+	} else if (content && typeof content == 'object') {
+		let elem = () => {
+			let elemm: JSX.Element = cloneWithProps(content, {
+				onClick: () => toggleActive(),
+			});
+			return elemm;
```
korgon, line 125: Can we just return cloneWithProps without the intermediary variable?

korgon (changes requested):

korgon (changes requested):
I think that removing the class name "ss__slideout__button" would cause some breaking changes.

packages/snap-preact-components/src/components/Molecules/Slideout/Slideout.stories.tsx
```diff
@@ -64,13 +64,13 @@ export default {
 			control: { type: 'text' },
 		},
 		buttonContent: {
-			defaultValue: 'click me',
+			defaultValue: '<div className="ss__slideout__button">click me</div>',
```
korgon, line 67: Should this be JSX now and not a string?

korgon: image

korgon: Looks like just removing this line solves things.

korgon (changes requested):
Not working in demo store...
image

korgon (approved):
Should be good to go.

PR #668, opened 2022-10-21
docs: add recommendations and github setup documentation

dkonieczek:


korgon (changes requested):
I think we will want to have a section for using the "Default Template" within the PREACT_RECOMMENDATIONS.md file. This default template will not require the use of Snapfu and ought to satisfy basic recommendations needs.

The default recommendation template for Snap includes a limited set of template parameters and will expect the component name to be `Default`.

I think we also may want to elaborate on the .json file and how to add additional custom template parameters.

Lastly, the instantiator config should probably be one of the first things in document prior to  discussion around templates - and the code example should include the `Default` component name in the mapping to align with the "Default Template" from the SMC.

docs/PREACT_RECOMMENDATIONS.md
```diff
@@ -1,12 +1,94 @@
...
+	const controller = props.controller;
+	const store = controller?.store;
+
+	if (!controller.store.loaded && !controller.store.loading) {
+		controller.search();
+	}
+
+	const parameters = store?.profile?.display?.templateParameters;
+
+	return store.results.length > 0 && <Recommendation controller={controller} />;
```
korgon, line 81: ^ there is currently a singular templateParameter `title` available for this template that we should be able to plug directly into the `Recommendation` component.

```
<Recommendation controller={controller} title={parameters.title} />
```

We need to ensure this works with the Default template - should be able to test tomorrow.

docs/PREACT_RECOMMENDATIONS.md
```diff
@@ -1,12 +1,94 @@
...
+});
+```
+
+## Custom Template
+Let's look at how to setup a custom recommendation template using the Snapfu CLI. See [Getting Started > Setup](https://searchspring.github.io/snap/#/start-setup) for installing Snapfu.
 
 There are three steps required for adding recommendations:
 - Creating the local template files
 - Uploading the newly created template to the Searchspring Management Console
```
korgon, line 90: Should this line be 'syncing' to match the snapfu command ?

korgon (changes requested):
Just a couple things in comments.
I like the new structure much better thanks!

korgon (approved):

PR #669, opened 2022-10-26
Facet Auto-open

chrisFrazier77:
adding a new config param to disable auto facet collapse handling

### DIFF
--- a/packages/snap-controller/src/Search/README.md
+++ b/packages/snap-controller/src/Search/README.md
@@ -13,6 +13,7 @@ The `SearchController` is used when making queries to the API `search` endpoint.
 | settings.facets.pinFiltered | selected facet options move to the top of the options array | true |   |
 | settings.facets.storeRange | range facets keep their initial boundary ranges from first load | true |   |
 | settings.facets.trim | facets that do not change results will be removed | true |   |
+| settings.facets.disableAutoCollapsedHandling | disable "auto open" functionality for facets that are filtered (active), collapsed, and have no stored data | false |   |
 | settings.facets.fields | object keyed by individual facet fields for configuration of any settings.facets options | ➖ |   |
 | settings.infinite | enable infinite scrolling by setting to empty object | ➖ |   |
 | settings.infinite.backfill | number of pages allowed for backfill | ➖ |   |
--- a/packages/snap-controller/src/Search/README.md
+++ b/packages/snap-controller/src/Search/README.md
@@ -13,7 +13,7 @@ The `SearchController` is used when making queries to the API `search` endpoint.
 | settings.facets.pinFiltered | selected facet options move to the top of the options array | true |   |
 | settings.facets.storeRange | range facets keep their initial boundary ranges from first load | true |   |
 | settings.facets.trim | facets that do not change results will be removed | true |   |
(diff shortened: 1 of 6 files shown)


packages/snap-store-mobx/src/types.ts
```diff
@@ -15,6 +15,7 @@ export type SearchStoreConfig = StoreConfig & {
 			singleResult?: boolean;
 		};
 		facets?: FacetStoreConfig & {
+			disableAutoCollapsedHandling?: boolean;
```
korgon, line 18: The configuration should not be here.

packages/snap-store-mobx/src/Search/Stores/SearchFacetStore.ts
```diff
@@ -251,13 +264,14 @@ export class ValueFacet extends Facet {
...
 		services: StoreServices,
 		storage: StorageStore,
 		facet: SearchResponseModelFacetValue,
-		facetMeta: MetaResponseModelFacet
+		facetMeta: MetaResponseModelFacet,
+		config: SearchStoreConfig | AutocompleteStoreConfig
 	) {
-		super(services, storage, facet, facetMeta);
+		super(services, storage, facet, facetMeta, config);
+		this.multiple = this.multiple;
```
korgon, line 274: Added a duplicated line here - should be removed.

packages/snap-store-mobx/src/Search/Stores/SearchFacetStore.ts
```diff
@@ -100,7 +100,8 @@ export class Facet {
 		services: StoreServices,
 		storage: StorageStore,
 		facet: SearchResponseModelFacetValue | SearchResponseModelFacetRange,
-		facetMeta: MetaResponseModelFacet
+		facetMeta: MetaResponseModelFacet,
+		config: SearchStoreConfig | AutocompleteStoreConfig
```
korgon, line 104: I think it makes more sense to pass FacetStoreConfig instead of the full store config here. It is already defined on line 49 above,  but would require some refactoring to get the non facet-specific configuration and default values set.

packages/snap-controller/src/Search/README.md
```diff
@@ -13,6 +13,7 @@ The `SearchController` is used when making queries to the API `search` endpoint.
 | settings.facets.pinFiltered | selected facet options move to the top of the options array | true |   |
 | settings.facets.storeRange | range facets keep their initial boundary ranges from first load | true |   |
 | settings.facets.trim | facets that do not change results will be removed | true |   |
+| settings.facets.disableAutoCollapsedHandling | disable "auto open" functionality for facets that are filtered (active), collapsed, and have no stored data | false |   |
```
korgon, line 16: I'm sure you saw this coming - but can we rename this?

disableAutoCollapsedHandling -> autoOpenActive

The default value should be true - but can be set to false via this config.

korgon (changes requested):
Few adjustments needed.

packages/snap-store-mobx/src/Search/Stores/SearchFacetStore.ts
chrisFrazier77: oops! nice catch

korgon (approved):

PR #675, opened 2022-10-28
Trending Terms Empty

chrisFrazier77:
dont save trending terms to storage if they are empty

also use the siteID in the storage key to allow switching of siteIds


korgon (approved):
Does the thing.

PR #683, opened 2022-11-07
Autocomplete history terms feature

chrisFrazier77:


packages/snap-controller/src/Autocomplete/AutocompleteController.ts
```diff
@@ -338,8 +338,13 @@ export class AutocompleteController extends AbstractController {
 			},
 			keyUp: (e: KeyboardEvent): void => {
 				// ignore enter and escape keys
-				if (e?.keyCode == KEY_ENTER || e?.keyCode == KEY_ESCAPE) return;
-
+				if (e?.keyCode == KEY_ENTER || e?.keyCode == KEY_ESCAPE) {
+					if (e.keyCode == KEY_ENTER) {
+						const input = e.target as HTMLInputElement;
```
dkonieczek, line 343: Maybe a question for @korgon 
Do you think the term that we save to history should be what was typed in the input or the suggested term? 
Or maybe only use the user typed in term if integrated spell correction is enabled?

packages/snap-preact-components/src/components/Organisms/Autocomplete/Autocomplete.tsx
```diff
@@ -170,6 +170,7 @@ export const Autocomplete = observer((properties: AutocompleteProps): JSX.Elemen
 		// default props
 		termsTitle: '',
 		trendingTitle: 'Popular Searches',
+		historyTitle: 'Previously Searched',
```
dkonieczek, line 173: `hideHistory` has a default value of `true` in the stories file however it looks to have a default value of `false` in the component file as it's not defined here.

packages/snap-preact-components/src/components/Organisms/Autocomplete/Autocomplete.tsx
```diff
@@ -454,7 +470,33 @@ export const Autocomplete = observer((properties: AutocompleteProps): JSX.Elemen
 												</div>
 											))}
 										</div>
-									</>
+									</div>
+								) : null}
+
+								{history.length && !hideHistory ? (
+									<div className="ss__autocomplete__terms__history">
```
dkonieczek, line 477: May want to not render the title divs if someone provides `<Autocomplete historyTitle={''}>`
Terms and Trending has this:
```
{trendingTitle ? (
    <div className="ss__autocomplete__title ss__autocomplete__title--trending">
        <h5>{trendingTitle}</h5>
    </div>
) : null}
```

packages/snap-preact-components/src/components/Organisms/Autocomplete/Autocomplete.tsx
```diff
@@ -454,7 +470,33 @@ export const Autocomplete = observer((properties: AutocompleteProps): JSX.Elemen
...
+										<div className="ss__autocomplete__title ss__autocomplete__title--history">
+											<h5>{historyTitle}</h5>
+										</div>
+										<div className="ss__autocomplete__terms__options ss__autocomplete__history__options">
+											{history.map((term: any) => (
```
dkonieczek, line 482: Can we remove the `any` type here to match `trending` and `terms` if it's not possible to use the `Term` type in all of these.

dkonieczek (changes requested):

packages/snap-controller/src/Autocomplete/AutocompleteController.ts
korgon: Yea, we should use the term (query) that we send with our redirect to the search page - in any case when using our autocomplete the first suggested term will contain the query that we direct to the search page with.

packages/snap-preact-components/src/components/Organisms/Autocomplete/Autocomplete.tsx
chrisFrazier77: good catch, it was true by default initially, but refactored to undefined.

korgon (changes requested):
It looks like we are only saving to history when the enter key is pressed - we should consider saving in all circumstances:
* when enter key is pressed
* when form is submitted
* when terms are clicked

Couple of thoughts:
Maybe instead of using the autocomplete controller/store to save history we should save them inside of the search controller? For example - when a search is run in a search controller we would save the query to history (perhaps on the SearchController search method execution). This would ensure that the history is accurate (would even build history via URLs). This could be done on a per siteId basis - and the AutocompleteController would simply instantiate a HistoryStore passing it a siteId in the config.

I'd like to see the history store be self contained and managed - by that I mean it should contain the methods to add to history, retrieve history or reset it. In instantiation it can have configuration for the limit and perhaps the storage name. This would make the history store more stand alone similar to other stores (pretty much just a wrapper around the storage store with special methods).

korgon (approved):

dkonieczek (approved):

PR #688, opened 2022-11-22
Carousel pagination prop bugfix

chrisFrazier77:


korgon (approved):
Looks like a good fix!

PR #690, opened 2022-11-23
Chunk Load Error Fix

chrisFrazier77:
adding a kill feature to snap for use with branch overrides, to prevent chunk errors


korgon (commented):
Seems like this could be tricky to test... Were you able to test this and verify that it indeed removes the branch override errors?

chrisFrazier77:
> Seems like this could be tricky to test... Were you able to test this and verify that it indeed removes the branch override errors?

little tricky indeed, but I used local overrides to change the script on the page to load localhost bundle and then had the branch param in the page url, which was picked up by the localhost bundle and that seemed to work as expected.

PR #693, opened 2022-12-12
Fix for `getBundleDetails()`

korgon:
Adjusting getBundleDetails to allow for local testing of branch override function.

The `!` was causing error because the details did not exist when forcing a branch override using local development server and browser override extension (to force to load local bundle for initial and branch bundle).


packages/snap-preact/src/getBundleDetails/getBundleDetails.ts
```diff
@@ -16,7 +16,7 @@ export const getBundleDetails = async (url: string): Promise<BundleDetails> => {
 				if ((lastModified && status === 0) || (status >= 200 && status < 400)) {
 					resolve({
 						url,
-						lastModified: lastModified!.split(',')[1].trim(),
+						lastModified: lastModified?.split(',')[1].trim() || '',
```
dkonieczek, line 19: 🤔 might need an optional chaining before the trim

dkonieczek (approved):

PR #699, opened 2022-12-16
feat: create snap-plugin package with bpk plugin

dkonieczek:


dkonieczek:
commit a new package-lock.json file after merging in https://github.com/searchspring/snap-generator-openapi/pull/26

PR #706, opened 2023-01-03
Re-enable Eslint 

korgon:
It seems eslint has been broken for a bit. There are so many errors and warnings now that I've run it successfully that we may want to create a new task to get this cleaned up.

We don't currently use eslint anywhere in the build pipeline, but maybe we ought to consider doing so. Without doing that we are unable to verify things like `debugger` are not in the code.


package.json
```diff
@@ -8,7 +8,7 @@
 	],
 	"scripts": {
 		"commit": "cz",
-		"build": "lerna exec --stream 'npm run build'",
+		"build": "lerna run build",
 		"build:docs": "lerna exec --stream --parallel 'npm run build:docs' && npm run build:feed",
```
dkonieczek, line 12: Do you think here we can also replace 
```lerna exec --stream --parallel 'npm run build:docs' && npm run build:feed``` 
with 
```lerna run build:docs && npm run build:feed```

As well as for `dev`, `lint` and `format`?

dkonieczek (approved):
Lerna 6, Nice!

package.json
korgon: That should not have been in this PR - sorry, wrong base selected...

But... I don't think that the build:docs command will benefit much from any of the caching though...

dkonieczek (approved):

PR #717, opened 2023-01-18
fix for redirects on single results after clearing filters

chrisFrazier77:


korgon (changes requested):
This is a nice clean fix!

In my testing I noticed that when clicking a hierarchy option from autocomplete it is still redirecting... This is because the response never includes hierarchies in the filters.

One way around this would be to check the request filters (and ignoring background filters)... maybe there is another way though?

korgon (commented):
Maybe add typing for request filter instead of any?

packages/snap-controller/src/Search/SearchController.ts
```diff
@@ -84,11 +84,13 @@ export class SearchController extends AbstractController {
 				return false;
 			}
 
+			let nonBackgroundFilters = search?.request?.filters?.filter((filter: any) => !filter.background);
```
korgon, line 87: Now that the request is typed, can we add a type for filter here?

packages/snap-controller/src/Search/SearchController.ts
```diff
@@ -84,11 +84,13 @@ export class SearchController extends AbstractController {
 				return false;
 			}
 
+			let nonBackgroundFilters = search?.request?.filters?.filter((filter) => !filter.background);
```
korgon, line 87: _Nitpicky_: If this isn't being re-assigned - `const`?

korgon (approved):
Types looking good!

korgon (approved):

PR #719, opened 2023-01-19
infinite scroll restoration enhancement

chrisFrazier77:


packages/snap-controller/src/Search/SearchController.ts
```diff
@@ -110,25 +110,49 @@ export class SearchController extends AbstractController {
...
-							window.scrollTo(0, scrollToPosition);
-							this.log.debug('scrolling to: ', scrollMap[stringyParams]);
-							window.clearInterval(heightCheck);
-						}
-						if (checkCount > 2000 / HEIGHT_CHECK_INTERVAL) {
-							window.clearInterval(heightCheck);
-						}
-						checkCount++;
-					}, HEIGHT_CHECK_INTERVAL);
+				if (this.config.settings?.infinite.selector) {
```
korgon, line 113: Maybe some of this code could be DRYed out a bit?

Lots of shared code here that I think could be simplified a bit.

korgon (changes requested):
Have you tested this out with the site that reported the issue?
In my testing with that site in `packages/snapps` it is not 100% working to restore the position - we should strive for a 100% success rate on restoration.

PR #722, opened 2023-01-26
adding a bool for personalized to merch store

chrisFrazier77:


korgon (changes requested):
Please add a test for this here as well:

https://github.com/searchspring/snap/blob/main/packages/snap-client/src/Client/transforms/searchResponse.test.ts

korgon (approved):

PR #724, opened 2023-01-28
Infinite Backfill Fix

korgon:
* fix duplication bug when infinite pages with over 500 results are backfilled

### DIFF
--- a/packages/snap-controller/src/Search/SearchController.ts
+++ b/packages/snap-controller/src/Search/SearchController.ts
@@ -287,6 +287,9 @@ export class SearchController extends AbstractController {
 						pageSize = meta.pagination?.defaultPageSize!;
 					}
 
+					// restricting pageSize to the limit
+					pageSize = pageSize > API_LIMIT ? API_LIMIT : pageSize;
+
 					const pagesNeeded =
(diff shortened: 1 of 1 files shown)


dkonieczek (approved):
Overall the normal use case looks good! Tested this on the demostore by setting results per page to 96 with a backfill of 15. Then when I'm on page 11 (1056 results) I'm seeing 3 requests each with `resultsPerPage: 352` instead of 500+500+56 like before. 

First edge case I found is that it looks like we don't restrict setting a value >500 for `globals.pagination.pageSize` ONLY when backfilling results. 

For example, if I set `globals.pagination.pageSize` to 600 then page 1 is as expected with 500 results because the api returned 500 results even though it's request params has `resultsPerPage: 600` so you see 1-500 results on page 1.

Then if you click the next pagination button to go to page 2 via the ajax request, you see 1-1000 results. But as soon as you refresh the page to backfill results, this changes to 1-1200. 

Another edge case I found is it looks like it requires an evenly divisible amount of results per page. I ran into this by setting `resultsPerPage: 510` while backfilling on page 3. 1530/4 is not a whole number so it splits it into 5 requests of 306.

dkonieczek (approved):

PR #733, opened 2023-02-03
Recommendation tracking components

chrisFrazier77:


korgon (changes requested):
Looking really good.

Couple of things:
* We should have a storybook entries for this.
* Readmes for these components should make it very clear that they should be used together.
* Since it is inheriting them, it should support all `ComponentProps`.
* Need testing around Tracker de-duping.
* May want to add testing around using different types of children in these components (to verify that `toChildArray` works as expected in all cases)

packages/snap-preact-components/src/components/Atoms/RecommendationProfileTracker/RecommendationProfileTracker.test.tsx
```diff
@@ -0,0 +1,416 @@
...
+
+const theme = {
+	components: {
+		RecommendationProfileTracker: {
+			className: 'themeClass',
+		},
+	},
+};
+
+describe('Recommendation Component', () => {
```
korgon, line 38: Should change text here to `RecommendationProfileTracker`.

packages/snap-preact-components/src/components/Atoms/RecommendationResultTracker/RecommendationResultTracker.stories.tsx
```diff
@@ -0,0 +1,78 @@
...
+	decorators: [
+		(Story: any) => (
+			<div style={{ maxWidth: '250px' }}>
+				<Story />
+			</div>
+		),
+	],
+	argTypes: {
+		controller: {
+			description: 'Controller reference',
```
korgon, line 34: Requires a specific `Recommendation` controller.

packages/snap-preact-components/src/components/Atoms/RecommendationResultTracker/RecommendationResultTracker.test.tsx
```diff
@@ -0,0 +1,331 @@
...
+
+const theme = {
+	components: {
+		RecommendationResultTracker: {
+			className: 'themeClass',
+		},
+	},
+};
+
+describe('Recommendation Component', () => {
```
korgon, line 37: Should change text to contain `RecommendationResultTracker`.

packages/snap-preact-components/src/components/Atoms/RecommendationResultTracker/RecommendationResultTracker.test.tsx
```diff
@@ -0,0 +1,331 @@
...
+			});
+		});
+
+		trackfn.mockClear();
+
+		const resultElem = rendered.container.querySelector('.findMe');
+
+		userEvent.click(resultElem!);
+
+		expect(trackfn).toHaveBeenCalled();
```
korgon, line 157: Should we verify what was clicked is what was tracked here?

packages/snap-preact-components/src/components/Atoms/RecommendationResultTracker/RecommendationResultTracker.tsx
```diff
@@ -0,0 +1,64 @@
...
+	};
+
+	const { children, result, controller, className, disableStyles, style } = props;
+
+	const resultRef = useRef(null);
+	const resultInViewport = useIntersection(resultRef, '0px');
+
+	if (resultInViewport) {
+		if (!controller.events.impression) {
+			controller.track.impression();
```
korgon, line 36: Should we be doing this?

I'm starting to think maybe this would encourage improper setups, or at the very least would lead to reports being partially correct which would make it hard to identify an invalid setup.

I have more to say on this in the main comment for the PR.

korgon (changes requested):
Please see comment throughout.

After splitting this up into two components (which I still think is the right way to do it) I am starting to think we may want to modify the RecommendationController tracking methods a bit... and this may not be important - but we may want to prevent tracking of result events when the profile events have not happened (which would indicate someone is using these trackers incorrectly). Maybe we should reject product render calls and impression calls if the profile has not first been tracked? Might need to think about this a bit more.

korgon (approved):

korgon (approved):

PR #754, opened 2023-02-21
ac component jank

chrisFrazier77:
refactor(autocomplete.tsx): refactor term display logic


korgon (commented):
Looks good.

Can you remove the similar jank when using `retainHistory` or `retainTrending` shown here:
janky

korgon (approved):
This is working well in all cases I tested.

PR #757, opened 2023-02-28
Scroll Event

korgon:
* adding `restorePosition` event and refactoring restoration code


korgon:
I'm starting to think that maybe this should be done regardless of infinite scroll or not...

chrisFrazier77 (approved):

PR #758, opened 2023-03-01
UrlManager Range Filters

chrisFrazier77:
url range filters bug fix when using tab manager

### DIFF
--- a/packages/snap-url-manager/src/Translators/Url/UrlTranslator.ts
+++ b/packages/snap-url-manager/src/Translators/Url/UrlTranslator.ts
@@ -176,14 +176,11 @@ export class UrlTranslator implements Translator {
 			.forEach((decodedHashEntries) => {
 				if (decodedHashEntries.length == 1) {
 					params.push({ key: [decodedHashEntries[0]], value: '', type: ParamLocationType.hash });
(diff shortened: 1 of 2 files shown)


packages/snap-url-manager/src/Translators/Url/UrlTranslator.ts
```diff
@@ -179,11 +179,17 @@ export class UrlTranslator implements Translator {
 				} else if (decodedHashEntries.length && decodedHashEntries.length <= 3) {
```
korgon, line 179: Is it possible to get rid of this condition and add additional logic in the <=4 block?

packages/snap-url-manager/src/Translators/Url/UrlTranslator.ts
```diff
@@ -361,6 +367,7 @@ export class UrlTranslator implements Translator {
 	}
 
 	serialize(state: UrlState): string {
+		// debugger
```
korgon, line 370: Not a big deal, but can we remove this line?

korgon (changes requested):
We should add a test for this in the UrlTranslator tests.

korgon (approved):

PR #761, opened 2023-03-06
Tag Version Upload

korgon:
* moving env variable assignment to a different step (apparently cannot use within step it is assigned)
* removing 'shell: bash' lines - as it is the default shell for Ubuntu runners
* adding a `clean:cache` to wipe the nx cache


.github/workflows/docs.yml
```diff
@@ -70,7 +74,6 @@ jobs:
           aws-region: us-east-1
 
       - name: Copy Files to S3
-        shell: bash
         run: |
           echo "current_tag=$(node .github/workflows/scripts/getTag.js --tags "$(git tag -l)")" >> $GITHUB_ENV
```
dkonieczek, line 78: should this line be removed?

dkonieczek (changes requested):

.github/workflows/docs.yml
korgon: Ubuntu default shell is `bash` - seems extraneous...

dkonieczek: I'm referring to the `echo "current_tag=` line here. It's duplicated from the addition above. 
I think an alternative solution would have also been to remove `>> $GITHUB_ENV` so that it sets the env variable at the shell and then could be referenced on the next lines via `$current_tag` instead of `${{ env.current_tag }}`

korgon: Oh yea, duplicated line... derp.

As far as the alternative, yea I considered that - but thought that having the current tag available to all steps could be beneficial in the future.

dkonieczek (dismissed):

dkonieczek (approved):

PR #764, opened 2023-03-09
Foreground Filtering

chrisFrazier77:
new feature to manually set foreground filters in search controller via config


packages/snap-url-manager/src/UrlManager/UrlManager.ts
```diff
@@ -29,7 +29,7 @@ class WatcherPool {
 export class UrlManager {
 	private translator: Translator;
 
-	private urlState: ImmutableObject<UrlState> = Immutable({}) as ImmutableObject<UrlState>;
+	urlState: ImmutableObject<UrlState> = Immutable({}) as ImmutableObject<UrlState>;
```
korgon, line 32: Any alternative to making public?

packages/snap-preact/src/setInitialUrlState/setInitialUrlState.ts
korgon, line 1: As discussed, we should probably not mix up the API request style with UrlManager state... Initial issue updated with suggested format for configuration.

korgon (changes requested):

korgon (changes requested):
After spending a bit of time with this I'm realizing there are some issues - for example it doesn't support custom state, nor does it handle all other possible initial states (like 'query' or 'tag').

Much of the logic being used I think would be automatically handled by using the merge function on the UrlManager itself.

I've got an idea on how we can simplify this code while also supporting full functionality - let's discuss further.

dkonieczek:
It looks like the foreground filters are affecting the page history.
If you search for a term (ie. shorts), the foreground filter is applied as expected. But if you press the back button, it only removes the foreground filters and leaves behind the query being searched instead of going back to the previous page.

Also noticed that it doesn't seem to apply the foreground filters when `integratedSpellCorrection` occurs. Try searching for 'blu' on the demo store to see this.

docs/INTEGRATION_FOREGROUND_FILTERS.md
```diff
@@ -0,0 +1,123 @@
+## Foreground Filters
+Foreground filters provide a way for pre-applying a filter on page load. The applied filter will be applied to the URL and can be removed as any other applied filter would. Foreground filtering is accomplished by setting the inital UrlManager state; this can be used for setting various states, but only filtering will be covered in this document.
```
dkonieczek, line 2: Consider renaming "covered in this document" to "covered in this example"?

docs/INTEGRATION_FOREGROUND_FILTERS.md
```diff
@@ -0,0 +1,123 @@
...
+
+Foreground filters are only usable with a SearchController.
```
dkonieczek, line 4: Consider bolding this text or move to the top to emphasize

docs/INTEGRATION_FOREGROUND_FILTERS.md
```diff
@@ -0,0 +1,123 @@
...
+			},
+		],
+	},
+};
+
+const snap = new Snap(config);
+```
+
+
+The `initial.parameters` object takes `UrlManager` state parameters, such as filter, sort, page, and pageSize. This feature is likely to mostly be used used with filter, but any parameters including custom ones could be utilized.
```
dkonieczek, line 41: Duplicate word - "used used" here: "This feature is likely to mostly be used used with filter,"
Also consider rephrasing to:
"This feature could be utilized by any parameter however is most likely to be used with the filter parameter"

docs/INTEGRATION_FOREGROUND_FILTERS.md
```diff
@@ -0,0 +1,123 @@
...
+	},
+};
+
+const snap = new Snap(config);
+```
+
+## Advanced Configuration
+Additional advanced configuration is available for special use cases. If you do not wish to use the default `ignoreParameters` you can specify as much using `initial.settings.useDefaultIgnoreParameters`, and setting it to `false`.
+
+More configuration can be made within Within each `initial.parameter` object. It is possible to specify individual `ignoreParameters` here, which allows for more control for when to apply each individual initial state. Within the parameter configuration it is also possible to opt out of the 'global' `ignoreParameter` settings specified in `initial.settings` by using the `useGlobalIgnoreParameters` configuration. Lastly, there may be cases where if isnecessary to replace existing state values instead of merging them (the default action). This is specified in each individual parameter configuration via the `action` configuration; this supports a `merge` or `set` value.
```
dkonieczek, line 83: Duplicate word - "within Within"

Rephrase "there may be cases where if isnecessary to" 
to "there may be cases where it is necessary to"

dkonieczek (changes requested):
Functionally looks good. Few documentation changes requested.

docs/INTEGRATION_FOREGROUND_FILTERS.md
korgon: All examples in the document only cover filtering - do you think it should still be re-worded?

docs/INTEGRATION_FOREGROUND_FILTERS.md
korgon: Agreed!

docs/INTEGRATION_FOREGROUND_FILTERS.md
dkonieczek: document sounds good

korgon (dismissed):

dkonieczek (dismissed):

korgon (approved):

PR #770, opened 2023-03-17
Lint

chrisFrazier77:


packages/snap-client/src/Client/apis/Recommend.ts
```diff
@@ -54,8 +54,7 @@ export class RecommendAPI extends API {
 	}
 
 	async batchRecommendations(parameters: RecommendRequestModel): Promise<RecommendResponseModel> {
-		let { tags, limits, categories, ...otherParams } = parameters;
-
+		let { tags, limits, categories, ...otherParams } = parameters; // eslint-disable-line
```
korgon, line 57: Can this be refactored to prevent the `eslint-disable-line` usage?

packages/snap-client/src/Client/apis/Recommend.ts
```diff
@@ -76,7 +75,7 @@ export class RecommendAPI extends API {
 			// now that the requests are in proper order, map through them
 			// and build out the batches
 			batch.entries.map((entry) => {
-				let { tags, limits, categories, ...otherParams } = entry.request;
+				let { tags, limits, categories, ...otherParams } = entry.request; // eslint-disable-line
```
korgon, line 78: Can this be refactored to prevent the `eslint-disable-line` usage?

.eslintrc.cjs
```diff
@@ -11,6 +11,27 @@ module.exports = {
 	rules: {
 		// add rules... or dont...
 		'no-debugger': 'error',
-		'@typescript-eslint/ban-ts-comment': 'off'
+		'@typescript-eslint/ban-ts-comment': 'off',
```
korgon, line 14: I think we should only allow with description:
https://github.com/typescript-eslint/typescript-eslint/blob/main/packages/eslint-plugin/docs/rules/ban-ts-comment.md#allow-with-description

.eslintrc.cjs
```diff
@@ -11,6 +11,27 @@ module.exports = {
...
+		'@typescript-eslint/no-explicit-any': 'off',
+		'@typescript-eslint/no-non-null-assertion': 'off',
```
korgon, line 16: `no-non-null-assertion` - I think this one should be enabled, as it can lead to run time errors if used incorrectly.

.eslintrc.cjs
```diff
@@ -11,6 +11,27 @@ module.exports = {
...
+		'@typescript-eslint/no-non-null-asserted-optional-chain': 'off',
```
korgon, line 17: `no-non-null-asserted-optional-chain` - similar to above, this one should also be enabled for similar reasons.

.eslintrc.cjs
```diff
@@ -11,6 +11,27 @@ module.exports = {
...
+		'@typescript-eslint/no-empty-function': 'off',
```
korgon, line 18: `no-empty-function` - we should consider enabling this as well, I suspect only our tests would be using empty functions, but if used in core code I'd question why they exist.

.eslintrc.cjs
korgon, line 1: `parserOptions.ecmaVersion` should probably be 2020 to match what we have in tsconfig:
https://github.com/searchspring/snap/blob/v0.42.3/tsconfig.json#L3-L4

.eslintrc.cjs
```diff
@@ -11,6 +11,27 @@ module.exports = {
...
+		"no-unused-vars": "off",
```
korgon, line 19: Is this redundant (or overwritten by the following rule)?

korgon (changes requested):
This is great - it is nice to see the linter actually implemented after all this time!

I tried adding an unused `const` variable, and it didn't catch it. I also tried adding a `debugger` line and it did not catch that. I think maybe something is not configured quite right?

---

One improvement I think we could make to speed up the linting time (at least locally) would be to run the linting in parallel by altering the lint script in the package.json:
`"lint": "lerna exec --parallel --stream 'npm run lint'",`

We could also maybe cache this (but probably not worthwhile since it goes so fast already)...

As a first iteration this is a good start. I have some comments on a few of the rules that I think we should change. I could see these as an additional task - but we could also just keep improving things right here.

.eslintrc.cjs
chrisFrazier77: I agree but I think this should be saved for a separate task, probably a big change

.eslintrc.cjs
chrisFrazier77: I agree but I think this should be saved for a separate task, probably a big change

chrisFrazier77: I agree but I think this should be saved for a separate task, probably a big change

.eslintrc.cjs
korgon: Yes can you please make an issue for this?

.eslintrc.cjs
korgon: Yes can you please make an issue for this?

korgon (approved):

PR #795, opened 2023-04-18
Display Settings Theme Merge Fix

korgon:
* changing the way that theme display settings are merged for breakpoints

### DIFF
--- a/packages/snap-preact-components/src/components/Organisms/Autocomplete/Autocomplete.tsx
+++ b/packages/snap-preact-components/src/components/Organisms/Autocomplete/Autocomplete.tsx
@@ -223,7 +223,7 @@ export const Autocomplete = observer((properties: AutocompleteProps): JSX.Elemen
 		controller?.setFocused && controller?.setFocused();
 	};
 
-	const themeOverride: Theme = {
+	const themeDefaults: Theme = {
 		components: {
 			facet: {
 				limit: 6,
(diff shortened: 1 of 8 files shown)


packages/snap-preact-components/src/components/Organisms/Autocomplete/Autocomplete.tsx
```diff
@@ -257,7 +257,7 @@ export const Autocomplete = observer((properties: AutocompleteProps): JSX.Elemen
 	};
 
 	const displaySettings = useDisplaySettings(breakpoints) || {};
-	const theme = deepmerge(themeOverride, deepmerge(props?.theme || {}, displaySettings?.theme || {}));
```
chrisFrazier77, line 260: looks like themeOverride is no longer used, I think this is still needed right? Maybe we need to do the same arrayMerge twice here? not sure what theming example you were using to test this all with, but that seems to pass tests and everything fine. 

`
const theme =  deepmerge(themeOverride, deepmerge(props?.theme || {}, displaySettings?.theme || {}, { arrayMerge: (destinationArray, sourceArray) => sourceArray }), { arrayMerge: (destinationArray, sourceArray) => sourceArray });
`

korgon: Oh good catch!

korgon: Corrected!

chrisFrazier77 (approved):

PR #796, opened 2023-04-18
refactor(autocomplete): handful of fixes and refactors regarding ac a…

chrisFrazier77:
…nd accontroller

### DIFF
--- a/packages/snap-preact-components/src/components/Organisms/Autocomplete/Autocomplete.tsx
+++ b/packages/snap-preact-components/src/components/Organisms/Autocomplete/Autocomplete.tsx
@@ -1,6 +1,6 @@
 /** @jsx jsx */
 import { h, Fragment } from 'preact';
-import { useEffect } from 'preact/hooks';
+import { useEffect, useState } from 'preact/hooks';
 
 import { observer } from 'mobx-react-lite';
 import { jsx, css } from '@emotion/react';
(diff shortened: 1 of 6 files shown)


packages/snap-preact-components/src/components/Organisms/Autocomplete/Autocomplete.tsx
```diff
@@ -74,6 +74,7 @@ const CSS = {
 				flexDirection: 'column',
 				flex: `1 1 auto`,
 				maxWidth: `${vertical || horizontalTerms ? 'auto' : '150px'}`,
+				minWidth: '150px',
```
korgon, line 77: I think this should be fine... But can you think of any negatives from introducing this min-width? I'm pretty sure we are preventing the render of `.ss__autocomplete__terms` when there are no terms right?

korgon (changes requested):
Looks good.
A few comments to clean up and a question for double checking on the style change.

packages/snap-preact-components/src/components/Organisms/Autocomplete/Autocomplete.tsx
chrisFrazier77: I _think_ it should be fine. I did a quick search in implementations repo for .ss__autocomplete__terms to see how they are using it, and didn't see any conflicting or alarming use cases. I pulled down a site and tested it and all looked good.

korgon (approved):

PR #803, opened 2023-04-25
Error Beacon refactor

chrisFrazier77:


korgon (commented):
I'm not feeling great about adding these extra details to `message`. I'm going to investigate the difficulty in adding a new column for this.

korgon:
@chrisFrazier77 
New columns have been added:
* context
* details

What I'd like to do is send contextual details to `context` instead of the `filename` column.

So instead of doing this:
https://github.com/searchspring/snap/blob/v0.43.1/packages/snap-controller/src/Abstract/AbstractController.ts#L69

We should instead send the context details as JSON data:
```
{
  controller: {
    id: this.id,
    type: this.type
  }
}
```

For now, I think controller is the only context to send - but in the future we could add more contextual information here.

The `details` should contain all the "additional details" about the error. In the case of our fetch errors, it would be the stuff that the client throws when it errors.

You should be able to test sending some new errors in this format to the Beacon, we can then inspect the database to ensure it made it in as expected.

packages/snap-controller/src/Abstract/AbstractController.test.ts
```diff
@@ -535,11 +535,22 @@ describe('Search Controller', () => {
...
+		controller.handleError(error, addedDetails);
 
 		expect(trackerTrackError).toHaveBeenCalledWith({
+			context: {
+				controller: {
+					type: 'abstract',
+					id: searchConfig.id,
+				},
+			},
+			details: JSON.stringify(addedDetails),
```
korgon, line 553: Not sure we have to stringify here (if we did, we'd need to also do the `context` above)... we should send some test errors to find out.

packages/snap-controller/src/Abstract/AbstractController.ts
```diff
@@ -66,12 +66,19 @@ export abstract class AbstractController {
...
 
 			const beaconPayload: TrackErrorEvent = {
-				filename: filename || `${this.id} (${this.type.charAt(0).toUpperCase() + this.type.slice(1)}Controller)`,
+				filename,
 				stack,
 				message,
 				colno,
 				lineno,
 				errortimestamp: timeStamp,
+				details: additionalDetails ? JSON.stringify(additionalDetails) : undefined,
```
korgon, line 75: As mentioned above (and against what was discussed elsewhere) - I don't think we need to do this stringify here. We should test.

packages/snap-tracker/src/types.ts
```diff
@@ -80,6 +80,10 @@ export interface BeaconContext {
 		type?: string;
 		id?: string;
 	};
+	controller?: {
+		type: string;
+		id: string;
+	};
```
korgon, line 86: This should be removed, as it is only passed into the `TrackErrorEvent` payload.

packages/snap-tracker/src/types.ts
```diff
@@ -94,14 +98,21 @@ export interface ShopperLoginEvent {
...
 	colno?: number;
 	lineno?: number;
 	errortimestamp?: number;
+	context?: {
+		controller?: {
+			type: string;
+			id: string;
+		};
+	};
+	details?: any;
```
korgon, line 115: Let's change the type to an object with keys of any type:
```
{
  [any: string]: unknown;
}
```

packages/snap-tracker/src/Tracker.ts
```diff
@@ -245,22 +245,23 @@ export class Tracker {
...
-			const { userAgent, href, filename, stack, message, colno, lineno, errortimestamp } = data;
+			const { useragent, href, filename, stack, message, colno, lineno, errortimestamp, details } = data;
 
 			const payload = {
 				type: BeaconType.ERROR,
 				category: BeaconCategory.RUNTIME,
 				context,
 				event: {
-					userAgent: userAgent || navigator.userAgent,
+					useragent: useragent || window?.navigator?.userAgent,
```
korgon, line 255: I forgot to mention in previous comments - but we can get rid of useragent altogether here, and we should, to prevent confusion.

We are now getting the user agent details from the beacon request themselves (on the beacon side of things).

korgon (changes requested):

korgon (approved):

PR #804, opened 2023-04-26
integrated Spell Correct fallbackQuery Change

chrisFrazier77:


korgon (changes requested):
Needs adjustments in `originalQuery` / `correctedQuery` logic - these should be dependent on `settings.integratedSpellCorrection`. Tests should not have needed altering - but new tests written to handle the change when using `settings.integratedSpellCorrection`.

korgon (approved):
Nice changes, this is exactly what we need.

PR #806, opened 2023-04-28
pageSizeOptions changes

chrisFrazier77:
adding optional configuration for pagination pageSizeOptions

### DIFF
--- a/packages/snap-controller/src/Search/README.md
+++ b/packages/snap-controller/src/Search/README.md
@@ -136,6 +136,51 @@ const searchConfig = {
 };
 ```
 
+
+### Pagination Settings 
+
+#### `pageSizeOptions` property
+The optional `pageSizeOptions` property gives the ability to overright the default pageSizeOptions in the searchPaginationStore. These are typically used in a `<select>` dropdown to change the number of results displayed per page. You can pass in as many options as you need. Each option has a `label` and a numeric `value`. The SearchPaginationStore then automatically populates and updates the `url` and `active` values. 
+
+`label` - label text to display
+
+`value` - number of results for this selection
+
+`url` - reference to the URLManager containing href and onclick values
+
+`active` - boolean stating if current page size matches the value of this option
+
+
+```typescript
+const searchConfig = {
+	id: 'search',
+	globals: {
+		pagination: {
+			pageSize: 12
+		}
+	},
+	settings: {
+		pagination: {
+			pageSizeOptions: [
+				{
+					label: `show 30`,
+					value: 30,
+				},
+				{
+					label: `show 40`,
+					value: 40,
+				},
+				{
+					label: `show 50`,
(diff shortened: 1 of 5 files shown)


packages/snap-store-mobx/src/Search/Stores/SearchPaginationStore.ts
```diff
@@ -37,22 +41,45 @@ export class SearchPaginationStore {
...
+				},
+			];
+
+			optionArr = defaultPageSizeOptions.map(
+				(pageOption: any) =>
+					new PageSizeOption(this.services, this.pageSize, {
+						label: pageOption.label,
+						value: pageOption.value,
+					})
+			);
```
korgon, line 76: This is identical to the map on line 48 above - can we consolidate/refactor this?

packages/snap-store-mobx/src/Search/Stores/SearchPaginationStore.ts
```diff
@@ -184,6 +211,29 @@ export class SearchPaginationStore {
 	}
 }
 
+export class PageSizeOption {
+	public services: StoreServices;
```
korgon, line 215: I think the services ought to be private - but I see it is not for the Page class currently... :(

packages/snap-controller/src/Search/README.md
```diff
@@ -136,6 +136,51 @@ const searchConfig = {
...
+
+
+```typescript
+const searchConfig = {
+	id: 'search',
+	globals: {
+		pagination: {
+			pageSize: 12
+		}
+	},
```
korgon, line 161: For this example, I think a pageSizeOption of 12 would make sense... but... doesn't this global `pageSize` overwrite the pageSizeOption selected? Not sure it should be here in the documentation if it does...

packages/snap-controller/src/Search/README.md
korgon, line 1: Totally unrelated, but I just noticed the `restorePosition` part of this documentation is inaccurate.

packages/snap-controller/src/Search/README.md
```diff
@@ -136,6 +136,51 @@ const searchConfig = {
 };
 ```
 
+
+### Pagination Settings 
+
+#### `pageSizeOptions` property
+The optional `pageSizeOptions` property gives the ability to overright the default pageSizeOptions in the searchPaginationStore. These are typically used in a `<select>` dropdown to change the number of results displayed per page. You can pass in as many options as you need. Each option has a `label` and a numeric `value`. The SearchPaginationStore then automatically populates and updates the `url` and `active` values.
```
korgon, line 143: searchPaginationStore -> SearchPaginationStore

korgon (changes requested):

korgon (approved):
Very nice!

PR #809, opened 2023-05-03
bugfix for emify

chrisFrazier77:
bugfix for autocomplete emify function when match was at the beginning of the term.


packages/snap-preact-components/src/components/Organisms/Autocomplete/Autocomplete.tsx
```diff
@@ -633,7 +633,7 @@ export const Autocomplete = observer((properties: AutocompleteProps): JSX.Elemen
 const emIfy = (term: string, search: string) => {
 	if (term && search) {
 		const match = term.match(escapeRegExp(search));
-		if (search && term && match && match.index) {
```
korgon, line 636: So was it the case that the index was _usually_ zero, and thus falsey here?

korgon (commented):
I tested using the branch demo store and it looks to be fixed, nice!
https://snapui.searchspring.io/snap/emify/index.html

korgon (approved):

PR #811, opened 2023-05-03
Instantiator Error Messages

chrisFrazier77:
bugfix for improper err logging when rec api call fails


packages/snap-preact/src/Instantiators/RecommendationInstantiator.tsx
```diff
@@ -248,6 +248,12 @@ export class RecommendationInstantiator {
 					await controller.search();
 				}
 
+				if (!controller.store.results.length && controller.store.error) {
+					//something went wrong
+					//err was already logged - nothing to do.
+					return;
```
korgon, line 254: I don't think we want to return here - as there are several things that happen after this. For example, adding the controller to the instantiator list of controllers, putting the controller in the window object, etc...

korgon (changes requested):

korgon (approved):

PR #829, opened 2023-05-30
sort search transform refactor

chrisFrazier77:
…
dont throw when incorrect params are passed to sort transform. instead log a warning and fallback to default sort and continue.

### DIFF
--- a/packages/snap-client/src/Client/transforms/searchRequest.ts
+++ b/packages/snap-client/src/Client/transforms/searchRequest.ts
@@ -29,12 +29,10 @@ transformSearchRequest.sorts = (request: SearchRequestModel = {}) => {
 		}
 
 		if (!sort.field || !sort.direction) {
-			console.warn('valid sort requires field and direction');
 			return {};
 		}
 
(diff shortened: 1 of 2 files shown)


packages/snap-client/src/Client/transforms/searchRequest.ts
korgon, line 1: I don't think console logs (or warnings belong here).

packages/snap-client/src/Client/transforms/searchRequest.ts
```diff
@@ -29,11 +29,11 @@ transformSearchRequest.sorts = (request: SearchRequestModel = {}) => {
 		}
 
 		if (!sort.field || !sort.direction) {
-			throw 'valid sort requires field and direction';
+			return {};
```
korgon, line 32: Looking at this code a bit closer I realize that you may need to return acc instead of the empty object here. Might want to take a closer look at this.

korgon (changes requested):

korgon (approved):

PR #836, opened 2023-06-02
Handle paused profile errors differently

chrisFrazier77:


packages/snap-tracker/src/Tracker.ts
```diff
@@ -265,7 +265,12 @@ export class Tracker {
 			};
 
 			// prevent sending of errors when on localhost or CDN
-			if (!payload.event.href || payload.event.href.includes('//localhost') || payload.event.href.includes('//snapui.searchspring.io/')) {
+			if (
+				payload.event.message?.indexOf('Profile is currently paused') !== -1 ||
```
dkonieczek, line 269: let's use `includes` to match the code style below

dkonieczek (changes requested):
Tested this on the demo store (commenting out the localhost check) and it no longer send an error beacon event as expected.
However it looks like the console still contains an error. Any way we could remove this and instead do a `log.warn` as mentioned in the issue?
Screenshot 2023-06-06 at 10 41 59 AM

packages/snap-tracker/src/Tracker.ts
chrisFrazier77: nice catch

PR #848, opened 2023-06-22
Release v0.45.1

### LINKED ISSUE #825: Slideout width
This is not a big issue, but very low cost to fix. I think the default ```width``` for slideout should be ```100%``` instead of ```90%```.
https://github.com/searchspring/snap/blob/main/packages/snap-preact-components/src/components/Molecules/Slideout/Slideout.tsx#L26

If you want a full width slideout and set the ```width``` prop to ```100%```, that doesn't achieve the effect. I do realize you can do a CSS override, but this seems like a small improvement easily made.

### LINKED ISSUE #826: Slideout docs
Testing in Chrome on Mac.

If you go to the Slideout section of the docs:
https://searchspring.github.io/snap/#/components-preact

You get a preview of the slideout where the slideout is open. If I switch to the "Docs" tab, I can't scroll to see the props. My guess is this has something to do with the slideout being fixed and overflow styles carrying over to this tab. 

To fix it, I can resize the screen on the "Docs" tab or close the slideout on the "Canvas" tab then go back to "Docs". Is it possible to close the slideout when clicking on the "Docs" tab?

### LINKED ISSUE #830: Prevent Recommendation Error
The `profile.json` API request has been changed to return a different message when the profile is being paused.

```
{
    "message": "Profile is currently paused"
}
```

https://8uyt2m.a.searchspring.io/api/personalized-recommendations/profile.json?tag=like&siteId=8uyt2m

When this happens in the RecommendationController, we should **_not_** send a error to the beacon (the current behavior). Instead, we should use the controller's logger to post a warning (`log.warn`) - warnings are not logged by default, but would be seen when using development mode (?dev).

### LINKED ISSUE #846: Update `snap-preact-demo` Cypress Version
Updating Cypress has many benefits - we are 4 main versions behind (currently at 8.5.0) and it is time that we update.

Cypress 12.15.0 is the latest.

dkonieczek:
closes #830 
closes #826 
closes #825 
closes #846


korgon:
Thanks @dkonieczek !

korgon (approved):

chrisFrazier77 (approved):

PR #859, opened 2023-08-16
preact-templates-typing

chrisFrazier77:
adding a place for custom result attribute typing to be added


korgon (approved):
This works well.

I think we may possibly want to re-name / re-organize in the future, but that will come later.

PR #865, opened 2023-08-17
terms component

chrisFrazier77:
clean up, testing, story, & documentation

### DIFF
--- a/packages/snap-preact-components/src/components/Atoms/Terms/Terms.test.tsx
+++ b/packages/snap-preact-components/src/components/Atoms/Terms/Terms.test.tsx
@@ -1,7 +1,7 @@
 import { h } from 'preact';
 
 import { v4 as uuidv4 } from 'uuid';
-import { render } from '@testing-library/preact';
+import { render, waitFor } from '@testing-library/preact';
 import { ThemeProvider } from '../../../providers';
 
 import { Terms } from './Terms';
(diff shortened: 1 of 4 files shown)


packages/snap-preact-components/src/components/Atoms/Terms/Terms.test.tsx
```diff
@@ -0,0 +1,221 @@
...
+	];
+
+	let controller: AutocompleteController;
+	beforeEach(async () => {
+		controller = createAutocompleteController({ client: clientConfig, controller: acConfig }, { client: mockClient });
+		await controller.bind();
+	});
+
+	it('renders', async () => {
+		// @ts-ignore - typing on terms
```
korgon, line 87: I'd like to see these ignores go away if possible. I know we've had typing issues in our tests for a long time. Was this due to the `'React' refers to a UMD global, but the current file is a module. Consider adding an import instead.ts(2686)` error?

It may be worth digging into it and trying to resolve fully for this and all the other tests where this is happening.

packages/snap-preact-components/src/components/Atoms/Terms/Terms.test.tsx
korgon, line 1: Seem to be missing a test for the `previewOnHover` prop.

Might be nice to have a few tests for non-standard usage - for example:

What should render when there are no terms?
What happens when you don't provide a controller or terms?
What happens when you provide a controller and terms (and they have different terms)?

korgon (changes requested):

korgon (approved):

PR #866, opened 2023-08-18
Initial theme

dkonieczek:


korgon (commented):
How can we disable a styleScript? For example a theme might be passing one in but we want it to be "un-themed". 
```
if (styleScript) {
  styling.css = [styleScript(stylingProps), style];
} else if (!disableStyles) {
  styling.css = [CSS.component(stylingProps), style];
} else if (style) {
  styling.css = [style];
}
```

dkonieczek:
> if (styleScript) {
>   styling.css = [styleScript(stylingProps), style];
> }

Perhaps this should also respect disableStyles? ie.
```
if (styleScript && !disableStyles) {
  styling.css = [styleScript(stylingProps), style];
} 
```

packages/snap-preact/src/Templates/themes/index.ts
```diff
@@ -1,14 +1,18 @@
 import { PikeTemplateTheme } from './pike';
+import { BocachicaTemplateTheme } from './bocachica';
 
 export type TemplateThemeConfig = {
-	name: keyof typeof themeMap;
+	name: 'pike' | 'bocachica'; // TODO: figure out why "keyof typeof themeMap" doesn't work here
```
korgon, line 5: Interesting, as this was seemingly working before...

korgon (changes requested):
Good start here!

When sending props to the styleScripts, I'd like to see if we can remove any additional properties that are not props whereever possible - this way there are no "mysterious" attributes that are not documented. I realize in some cases this will not be possible, but where able, we should clean this up.

I did not spend much time looking closely at the styleScripts in Bocachica this time around, as I feel like the variable usage will be in flux until we settle on the schema.

packages/snap-preact-components/src/components/Atoms/Loading/LoadingBar.tsx
```diff
@@ -44,21 +44,25 @@ const CSS = {
...
 
 	const props = mergeProps('loadingBar', globalTheme, defaultProps, properties);
 
-	const { active, color, backgroundColor, height, disableStyles, className, style } = props;
+	const { active, disableStyles, className, style, styleScript } = props;
 
 	const styling: { css?: StylingCSS } = {};
-	if (!disableStyles) {
-		styling.css = [CSS.loadingBar({ height, color, backgroundColor, theme, animation: CSS.animation }), style];
+	const stylingProps = { ...props, animation: CSS.animation };
```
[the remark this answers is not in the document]
dkonieczek: Agreed. I'll remove this and instead move the keyframes to the styleScript itself.

packages/snap-preact-components/src/components/Atoms/Loading/LoadingBar.tsx
```diff
@@ -44,21 +44,25 @@ const CSS = {
...
 
 	const props = mergeProps('loadingBar', globalTheme, defaultProps, properties);
 
-	const { active, color, backgroundColor, height, disableStyles, className, style } = props;
+	const { active, disableStyles, className, style, styleScript } = props;
 
 	const styling: { css?: StylingCSS } = {};
-	if (!disableStyles) {
-		styling.css = [CSS.loadingBar({ height, color, backgroundColor, theme, animation: CSS.animation }), style];
+	const stylingProps = { ...props };
```
korgon, line 56: Should we maybe find all places where we are doing this and just set `stylingProps = props` instead of spreading into a new object? Not a huge difference but when you are doing this across hundreds of components it could add up.   ... maybe...

packages/snap-preact-components/src/components/Atoms/Button/Button.stories.tsx
```diff
@@ -66,6 +67,18 @@ export default {
 			},
 			control: { type: 'color' },
 		},
+		icon: {
+			description: 'Button Icon name',
```
korgon, line 71: Yea, we might want this in Storybook eh?  ;)

Nice catch.

korgon (changes requested):
Looks like we've got merch conflicts... :|

packages/snap-preact-components/src/components/Atoms/Button/Button.stories.tsx
dkonieczek: I think this came from Chris so props to him for this catch! https://github.com/searchspring/snap/commit/92bdb99c25955b17995671d6475697534152ee1d#diff-af43f9ca968d24cd962c27d74c9fa94a839d9094bbe15052a237d341fd9754c3R71

korgon (approved):

PR #870, opened 2023-08-22
Container component

chrisFrazier77:
moving flex to container and refactoring. added tests


packages/snap-preact-components/src/components/Layouts/utils/Container/Container.tsx
```diff
@@ -0,0 +1,246 @@
...
+		| 'revert-layer'
+		| 'unset';
+	flex?: string | number;
+	flexBasis?: string | number;
+	textAlign?: 'left' | 'right' | 'center' | 'justify' | 'initial' | 'inherit';
+	gap?: string | number;
+	rowGap?: string | number;
+	columnGap?: string | number;
+	item?: boolean;
+	href?: string;
```
korgon, line 245: Since this is a clickable thing that clients may want to tie into, do you think we should add an `onClick` prop?

korgon (changes requested):
It is evolving! Nice changes.

packages/snap-preact-components/src/components/Layouts/utils/Container/Container.tsx
chrisFrazier77: Wouldn't that open the door for some unwanted crazy customization?

korgon: Currently the Button component is needed to do `onClick` functionality. My thinking here was that having a link may also require some additional tracking, and there is no possible way to handle that currently (outside of the Button component).

korgon: Let's leave it off for now until use case arises.

korgon (approved):

PR #872, opened 2023-08-23
Preact Components - Element

chrisFrazier77:
converting string - element, adding tests and story and docs


packages/snap-preact-components/src/components/Atoms/Element/Element.stories.tsx
```diff
@@ -0,0 +1,81 @@
...
+			},
+			control: {
+				type: 'object',
+			},
+		},
+		...componentArgs,
+	},
+};
+
+export const Default = (args: ElementProps) => <Element {...args} />;
```
korgon, line 61: Let's just name this one Div.

packages/snap-preact-components/src/components/Atoms/Element/Element.stories.tsx
```diff
@@ -0,0 +1,81 @@
...
+	content: 'content text',
+};
+
+export const Span = (args: ElementProps) => <Element {...args} />;
+Span.args = {
+	type: 'span',
+	content: "I'm a span",
+};
+
+export const Attributes = (args: ElementProps) => <Element {...args} />;
```
korgon, line 73: Let's name this one Label

packages/snap-preact-components/src/components/Atoms/Element/Element.stories.tsx
```diff
@@ -0,0 +1,81 @@
...
+Attributes.args = {
+	type: 'label',
+	content: "I'm a label with attributes",
+	attributes: {
+		attribute1: 'hi',
+		attribute2: 'mom',
```
korgon, line 79: I think we ought to use something a little more professional here.
Maybe some data attributes or something similar?

packages/snap-preact-components/src/components/Atoms/Element/Element.tsx
```diff
@@ -0,0 +1,61 @@
...
+
+	return type ? (
+		<CacheProvider>
+			{type == 'div' && <div {...args}></div>}
+
+			{type == 'span' && <span {...args}></span>}
+
+			{type == 'p' && <p {...args}></p>}
+
+			{type == 'label' && <label {...args}></label>}
```
korgon, line 46: I like this approach, nice.

packages/snap-preact-components/src/components/Atoms/Element/readme.md
```diff
@@ -0,0 +1,33 @@
+# Element
+
+Renders an html element of your choice. This component can be used for adding basic strings to your layouts, or for adding specific elements for third parties to target and inject into.
```
korgon, line 3: Can we capitalize HTML here?

packages/snap-preact-components/src/components/Layouts/ResultLayout/ResultLayout.tsx
```diff
@@ -77,7 +77,7 @@ export type LayoutElement = {
 	| PriceElement
 	| RatingElement
 	| SkeletonElement
-	| StringElement
+	| ElementElement
```
korgon, line 80: 😁

packages/snap-preact-components/src/components/Organisms/Sidebar/Sidebar.tsx
```diff
@@ -101,7 +101,7 @@ export const Sidebar = observer((properties: SidebarProps): JSX.Element => {
 	return (
 		<CacheProvider>
 			<div {...styling} className={classnames('ss__sidebar')}>
-				{!hideTitle && <String className="ss--sidebar-title" content={subProps.String.content as string} {...subProps.String} />}
+				{!hideTitle && <Element className="ss--sidebar-title" type={'span'} {...subProps.Element} />}
```
korgon, line 104: Any reason we are using Element here? Could we just use an H tag instead?

packages/snap-preact-components/src/components/Atoms/Element/Element.tsx
```diff
@@ -0,0 +1,61 @@
...
+	);
+});
+
+export interface ElementProps extends ComponentProps {
+	type: 'div' | 'span' | 'p' | 'label';
+	content?: string;
+	attributes?: {
+		[name: string]: string | number | undefined;
+		//no onclicks allowed
+		onClick?: undefined;
```
korgon, line 59: I'm wondering if this preventing onClick is necessary here...
And if so, do we then need to prevent all types of other events as well?

korgon (changes requested):

packages/snap-preact-components/src/components/Atoms/Element/Element.tsx
chrisFrazier77: This came from our discussion previously about all the bad things someone can do with onclick functions. removing the ability removed custom code. Things with onClicks are not what this component is intended for. And by allowing it will cause people to abuse it and run wild creating their own "components". At least thats what we had decided previously. 

As for the other events, I had a similar thought 🤷  I could see them being dangerous as well.

packages/snap-preact-components/src/components/Organisms/Sidebar/Sidebar.tsx
korgon: Yea, I think the usage here should probably be removed, since it is unnecessary.

packages/snap-preact-components/src/components/Atoms/Element/Element.tsx
korgon: Yea, I remember that discussion, but I'm just not sure there is much point in preventing it here when there are other ways of doing it elsewhere. Also, how many other events would be have to try and prevent here?

packages/snap-preact-components/src/components/Organisms/Sidebar/Sidebar.tsx
```diff
@@ -101,7 +90,7 @@ export const Sidebar = observer((properties: SidebarProps): JSX.Element => {
 	return (
 		<CacheProvider>
 			<div {...styling} className={classnames('ss__sidebar')}>
-				{!hideTitle && <Element className="ss--sidebar-title" type={'span'} {...subProps.Element} />}
+				{!hideTitle && <h4 className="ss--sidebar-title">{title}</h4>}
```
korgon, line 93: `ss__sidebar__title`

korgon (changes requested):

packages/snap-preact-components/src/components/Atoms/Element/Element.tsx
korgon: Let's revisit the need for `onClick` and potentially preventing other events after we experiment more.

PR #881, opened 2023-08-31
Mobile sidebar

chrisFrazier77:


packages/snap-preact-components/src/components/Organisms/MobileSidebar/MobileSidebar.tsx
korgon, line 1: I like how you have separated the header, body and footer content (title, sidebar, cta-wrapper), but I think we ought to change these class names:

`ss__mobile-sidebar__title` => `ss__mobile-sidebar__header`
`ss__mobile-sidebar__sidebar` => `ss__mobile-sidebar__body`
`ss__mobile-sidebar__cta-wrapper` => `ss__mobile-sidebar__footer` 

Other changes I think would be good would be changing the prop names around a bit, starting with the header and footer.

`hideHeader` (instead of 'hideTitle')
`hideFooter` (no equivalent exists currently)

I think some consistency in naming could be helpful for some of the other props as well. For example `clearButtonText` and `hideClearAllButton` are referring to the same thing, but are named slightly differently. Additionally, for some of these buttons, an icon might be nice to have.

`openButtonText` (instead of 'slideoutButtonText')
`openButtonIcon` (instead of 'iconOpen')

`closeButtonText` (doesn't currently exist, but seems reasonable to want 'close' instead of icon here)
`closeButtonIcon` (instead of 'iconClose')
`hideCloseButton`

`clearButtonText`
`clearButtonIcon` (doesn't currently exist, but seems reasonable to want an icon and no text or both)
`hideClearButton` (instead of 'hideClearAllButton')

`applyButtonText`
`applyButtonIcon` (doesn't currently exist, but seems reasonable to want an icon and no text or both)
`hideApplyButton`

packages/snap-preact-components/src/components/Organisms/MobileSidebar/MobileSidebar.tsx
```diff
@@ -0,0 +1,219 @@
...
+		},
+	};
+
+	const Content = (props: any) => {
+		const { toggleActive } = props;
+		return (
+			<div className="ss__mobile-sidebar__content">
+				{!hideTitle && (
+					<div className="ss__mobile-sidebar__title">
+						<span>{titleText}</span>
```
korgon, line 134: I think we may want to have a classname on this element (and maybe use an H tag?) - eg: `ss__mobile-sidebar__header__title`

packages/snap-preact-components/src/components/Organisms/MobileSidebar/MobileSidebar.test.tsx
korgon, line 1: Missing tests for button functionality - specifically the apply and clear buttons in the footer.

packages/snap-preact-components/src/components/Organisms/MobileSidebar/readme.md
korgon, line 1: Will need changes after standardization of prop names.

packages/snap-preact-components/src/components/Organisms/Sidebar/Sidebar.stories.tsx
```diff
@@ -0,0 +1,128 @@
...
+					summary: 'string',
+				},
+				defaultValue: { summary: 'Filters' },
+			},
+			control: { type: 'text' },
+		},
+
+		hideTitle: {
+			defaultValue: false,
+			description: 'hides the sidebar title component',
```
korgon, line 57: Not _really_ a component...  /shrug

packages/snap-preact-components/src/components/Organisms/Sidebar/Sidebar.stories.tsx
```diff
@@ -0,0 +1,128 @@
...
+				type: {
+					summary: 'boolean',
+				},
+				defaultValue: { summary: false },
+			},
+			control: { type: 'boolean' },
+		},
+		hidePerPage: {
+			defaultValue: false,
+			description: 'hides the sidebar per page component',
```
korgon, line 79: per page -> PerPage

packages/snap-preact-components/src/components/Organisms/Sidebar/Sidebar.stories.tsx
```diff
@@ -0,0 +1,128 @@
...
+				type: {
+					summary: 'boolean',
+				},
+				defaultValue: { summary: false },
+			},
+			control: { type: 'boolean' },
+		},
+		hideSortBy: {
+			defaultValue: false,
+			description: 'hides the sidebar sort by component',
```
korgon, line 90: sort by -> SortBy

packages/snap-preact-components/src/components/Organisms/Sidebar/Sidebar.tsx
```diff
@@ -89,16 +89,16 @@ export const Sidebar = observer((properties: SidebarProps): JSX.Element => {
 
 	return (
 		<CacheProvider>
-			<div {...styling} className={classnames('ss__sidebar')}>
-				{!hideTitle && <h4 className="ss__sidebar__title">{title}</h4>}
+			<div {...styling} className={classnames('ss__sidebar', className)}>
+				{!hideTitle && <h4 className="ss__sidebar__title">{titleText}</h4>}
```
korgon, line 93: This shown when the page is loading, might want to wrap the entire thing in some check?

packages/snap-preact-components/src/components/Templates/Search/Search.tsx
```diff
@@ -158,12 +171,18 @@ export const Search = observer((properties: SearchProps): JSX.Element => {
...
+
 	return (
 		<ThemeProvider theme={properties.theme || {}}>
 			<CacheProvider>
 				<div {...styling} className={classnames('ss__search', className)}>
-					{!hideSidebar && <Sidebar {...subProps.Sidebar} controller={controller} />}
+					<div className="ss__search__sidebar-wrapper">
+						{!hideSidebar && !isMobile && <Sidebar {...subProps.Sidebar} controller={controller} />}
 
+						<MobileSidebar controller={controller} {...subProps.MobileSidebar} />
```
korgon, line 184: The positioning here is awkward and could be better - but something we can adjust later.

image

packages/snap-preact-components/src/components/Organisms/MobileSidebar/MobileSidebar.stories.tsx
```diff
@@ -0,0 +1,227 @@
...
+			},
+		},
+		...componentArgs,
+	},
+};
+
+const snapInstance = Snapify.search({ id: 'Facet', globals: { siteId: '8uyt2m' } });
+
+export const Default = (args: MobileSidebarProps, { loaded: { controller } }: { loaded: { controller: SearchController } }) => {
+	return <MobileSidebar {...args} controller={controller} />;
```
korgon, line 217: For this default example in storybook, I think changing the openButtonText could be helpful; maybe something like 'click me' or 'click to open filters' or similar to make it more obvious.

korgon (changes requested):
Looking good - few adjustments in comments.

packages/snap-preact-components/src/components/Organisms/Sidebar/Sidebar.tsx
korgon, line 1: Thinking about things... Should the left Banner component be here? Perhaps a prop to hide it as well?

korgon (changes requested):
With the changes to the Button component to include an icon, we will possibly want/need a wrapper around the button text. Without the wrapper it is impossible to use flex ordering on the button contents (eg. want to make the icon first). Although, I noticed that using an `order: -1` on the icon element does this also... However, styling just this text (without affecting the icon) would be more difficult without a wrapper.

packages/snap-preact-components/src/components/Organisms/Sidebar/Sidebar.tsx
chrisFrazier77: TBD

korgon (approved):

PR #882, opened 2023-09-05
No results

chrisFrazier77:


packages/snap-preact-components/src/components/Atoms/NoResults/NoResults.stories.tsx
```diff
@@ -0,0 +1,111 @@
...
+	decorators: [
+		(Story: any) => (
+			<div style={{ maxWidth: '250px' }}>
+				<Story />
+			</div>
+		),
+	],
+	argTypes: {
+		staticSlot: {
+			description: 'Slot for more overriding no results content entirely',
```
korgon, line 31: Remove "more" word?

packages/snap-preact-components/src/components/Atoms/NoResults/NoResults.stories.tsx
korgon, line 1: Need a story for DYM use case.

packages/snap-preact-components/src/components/Atoms/NoResults/NoResults.tsx
```diff
@@ -0,0 +1,132 @@
...
+
+const CSS = {
+	noresults: () => css({}),
+};
+
+export const NoResults = observer((properties: NoResultsProps): JSX.Element => {
+	const globalTheme: Theme = useTheme();
+
+	const defaultProps: Partial<NoResultsProps> = {
+		dymText: `Did you mean <a href=${properties.controller?.store?.search?.didYouMean?.url.href}>${properties.controller?.store?.search?.didYouMean?.string}</a>?`,
```
korgon, line 22: Seems like this should maybe be a function? Otherwise how would the text be set properly?

But at the same time - does this really belong here? Or should it really be in the `SearchHeader`?

packages/snap-preact-components/src/components/Atoms/NoResults/NoResults.tsx
```diff
@@ -0,0 +1,132 @@
...
+	const styling: { css?: StylingCSS } = {};
+	if (!disableStyles) {
+		styling.css = [CSS.noresults(), style];
+	} else if (style) {
+		styling.css = [style];
+	}
+
+	return (
+		<div className={classnames('ss__no-results', className)}>
+			{staticSlot ? (
```
korgon, line 77: I'm not sure the static slot should exclude usage of the other props...

packages/snap-preact-components/src/components/Atoms/NoResults/NoResults.tsx
```diff
@@ -0,0 +1,132 @@
...
+										))}
+									</ul>
+								) : null}
+							</>
+						)}
+					</div>
+
+					{!hideContact && contactsList && contactsList.length !== 0 && (
+						<div className="ss__no-results__contact">
+							{contactsTitleText && <p className="ss__no-results__contact__title">{contactsTitleText}</p>}
```
korgon, line 101: The default value for `contactsTitleText` has HTML but it is not rendered here.

packages/snap-preact-components/src/components/Atoms/NoResults/NoResults.tsx
korgon: The slot should probably have a wrapper div to handle changing the order via CSS as needed.

packages/snap-preact-components/src/components/Atoms/NoResults/NoResults.tsx
```diff
@@ -0,0 +1,132 @@
...
+		styling.css = [style];
+	}
+
+	return (
+		<div className={classnames('ss__no-results', className)}>
+			{staticSlot ? (
+				<Fragment>{cloneWithProps(staticSlot, { controller })}</Fragment>
+			) : (
+				<Fragment>
+					{dym && <p className="ss__no-results__did-you-mean" dangerouslySetInnerHTML={{ __html: dymText! }}></p>}
```
korgon, line 81: As mentioned above, this maybe belongs in the `SearchHeader` component?

packages/snap-preact-components/src/components/Atoms/NoResults/NoResults.tsx
```diff
@@ -0,0 +1,132 @@
...
+
+					{!hideContact && contactsList && contactsList.length !== 0 && (
+						<div className="ss__no-results__contact">
+							{contactsTitleText && <p className="ss__no-results__contact__title">{contactsTitleText}</p>}
+
+							{contactsList.map((contact: any) => (
+								<div className={`ss__no-results__contact__detail ss__contact__detail--${handleize(contact.title)}`}>
+									<h4 className="ss__no-results__contact__detail__title">{contact.title}</h4>
+
+									<p dangerouslySetInnerHTML={{ __html: contact.content }}></p>
```
korgon, line 107: Should probably have a class on here?

packages/snap-preact-components/src/components/Atoms/NoResults/NoResults.tsx
```diff
@@ -0,0 +1,132 @@
...
+
+export interface NoResultsProps extends ComponentProps {
+	staticSlot?: any;
+	dymText?: string;
+	suggestionsTitleText?: string;
+	suggestionsList?: string[];
+	hideContact?: boolean;
+	contactsTitleText?: string;
+	contactsList?: contact[];
+	controller?: SearchController;
```
korgon, line 131: If we got rid of `dymText` I suppose we could get rid of this prop eh?

korgon (changes requested):
Few things, biggest being about the DYM.

korgon (approved):

PR #886, opened 2023-09-11
Ratings component

chrisFrazier77:
Testing is still needed

### DIFF
--- a/packages/snap-preact-components/src/components/Molecules/Rating/Rating.tsx
+++ b/packages/snap-preact-components/src/components/Molecules/Rating/Rating.tsx
@@ -94,12 +94,12 @@ export const Rating = observer((properties: RatingProps): JSX.Element => {
 	}
 
 	const styling: { css?: StylingCSS } = {};
-	const stylingProps = { ...props };
+	const stylingProps = { ...props, value };
 
 	if (styleScript && !disableStyles) {
 		styling.css = [styleScript(stylingProps), style];
 	} else if (!disableStyles) {
-		styling.css = [CSS.rating({ ...props, value }), style];
+		styling.css = [CSS.rating(stylingProps), style];
 	} else if (style) {
 		styling.css = [style];
 	}
--- a/packages/snap-preact-components/src/components/Molecules/Rating/Rating.tsx
+++ b/packages/snap-preact-components/src/components/Molecules/Rating/Rating.tsx
@@ -6,32 +6,40 @@ import { jsx, css } from '@emotion/react';
 import classnames from 'classnames';
 import { Theme, useTheme, CacheProvider } from '../../../providers';
 import { ComponentProps, StylingCSS } from '../../../types';
-import { mergeProps } from '../../../utilities';
+import { defined, mergeProps } from '../../../utilities';
+import { Icon, IconProps } from '../../Atoms/Icon';
 
 const CSS = {
-	rating: (emptyRatingSrc?: string, fullRatingSrc?: string) =>
+	rating: () =>
 		css({
-			textAlign: 'left',
-			height: '24px',
-			margin: '10px auto',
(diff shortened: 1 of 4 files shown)


packages/snap-preact-components/src/components/Molecules/Rating/Rating.tsx
```diff
@@ -6,84 +6,155 @@ import { jsx, css } from '@emotion/react';
...
 
 	const styling: { css?: StylingCSS } = {};
-	if (!disableStyles) {
-		styling.css = [CSS.rating(emptyRatingSrc, fullRatingSrc), style];
+	const stylingProps = { ...props };
+
+	if (styleScript && !disableStyles) {
+		styling.css = [styleScript(stylingProps), style];
+	} else if (!disableStyles) {
+		styling.css = [CSS.rating({ ...props, value }), style];
```
dkonieczek, line 102: Value isn't needed in the default component styles? But might be fine to have it here if it's dereferenced from `props.value`

packages/snap-preact-components/src/components/Molecules/Rating/Rating.tsx
```diff
@@ -6,84 +6,155 @@ import { jsx, css } from '@emotion/react';
...
+		value = 0;
+	}
+	if (value > 5) {
+		value = 5;
+	}
 
 	const styling: { css?: StylingCSS } = {};
-	if (!disableStyles) {
-		styling.css = [CSS.rating(emptyRatingSrc, fullRatingSrc), style];
+	const stylingProps = { ...props };
```
dkonieczek, line 97: Should `value` be added here so that it can be accessed in the styleScript and so that it overwrites `props.value` since it's a primitive and would looses reference if the above if statements change it's value?

packages/snap-preact-components/src/components/Molecules/Rating/Rating.tsx
```diff
@@ -6,84 +6,155 @@ import { jsx, css } from '@emotion/react';
...
-	if (isNaN(rating)) {
-		rating = 0;
-	}
-	if (rating > 5) {
-		rating = 5;
-	}
-
-	const star = Math.floor(rating * 20);
+	// with 'disablePartialFill' we are rounding down
+	const numStarsToShow = disablePartialFill ? Math.floor(value) : Math.ceil(value);
```
dkonieczek, line 108: With `disablePartialFill` disabled, `Math.ceil(4.1)` will give you `5`. Do you think using `Math.round` would be better? Or should this be another prop?

korgon: So the `numStarsToShow` is expected to be 5 when we are doing partial fill. It then will calculate the last star width in the icon map below (using `numStarsToShow` to determine how many _total_ icons to render).

packages/snap-preact-components/src/components/Molecules/Rating/Rating.tsx
korgon: Ah yes, I will update stylingProps accordingly.

packages/snap-preact-components/src/components/Molecules/Rating/Rating.tsx
korgon: Ah yes, I will update stylingProps accordingly.

dkonieczek (approved):

packages/snap-preact-components/src/components/Molecules/Rating/Rating.tsx
korgon: If we need to add a new prop to determine the rounding to be used when using `disablePartialFill` (currently we only round down via "floor") we could do that at a future date.

PR #900, opened 2023-09-27
search template

chrisFrazier77:


packages/snap-preact-components/src/components/Templates/Search/Search.test.tsx
```diff
@@ -0,0 +1,318 @@
...
+	it('can hide hideSearchHeader', () => {
+		const rendered = render(<Search controller={controller} hideSearchHeader />);
+		const element = rendered.container.querySelector('.ss__search')!;
+		const header = rendered.container.querySelector('.ss__search-header');
+
+		expect(element).toBeInTheDocument();
+		expect(header).not.toBeInTheDocument();
+	});
+
+	it('can hide toptoolbar', () => {
```
korgon, line 110: topToolbar

packages/snap-preact-components/src/components/Templates/Search/Search.test.tsx
```diff
@@ -0,0 +1,318 @@
...
+		const rendered = render(<Search controller={controller} />);
+
+		const element = rendered.container.querySelector('.ss__search');
+		const banners = rendered.container.querySelectorAll('.ss__banner');
+		const headerBanner = rendered.container.querySelector('.ss__banner--header');
+		const bannerBanner = rendered.container.querySelector('.ss__banner--banner');
+		const footerBanner = rendered.container.querySelector('.ss__banner--footer');
+
+		expect(element).toBeInTheDocument();
+		expect(banners).toHaveLength(3);
```
korgon, line 209: Should be length 4 on desktop, as the `left` banner would be visible.

packages/snap-preact-components/src/components/Templates/Search/Search.test.tsx
```diff
@@ -0,0 +1,318 @@
...
+		const banners = rendered.container.querySelectorAll('.ss__banner');
+
+		expect(element).toBeInTheDocument();
+		expect(banners).toHaveLength(0);
+	});
+
+	it('can hide only some merchandising banners', async () => {
+		await controller.search();
+
+		const rendered = render(<Search controller={controller} hideMerchandisingBanners={[ContentType.FOOTER, ContentType.HEADER]} />);
```
korgon, line 230: Do we have to use enums here? Just curious how this would work in the integration.

packages/snap-preact-components/src/components/Templates/Search/readme.md
```diff
@@ -0,0 +1,84 @@
...
+
+### hideMerchandisingBanners
+The `hideMerchandisingBanners` prop specifies if merchandising banners should render. This can take a boolean to hide all banners, or an array of specific banner types you wish to hide. 
+
+```jsx
+<Search controller={controller} hideMerchandisingBanners={true} />
+```
+or
+```jsx
+<Search controller={controller} hideMerchandisingBanners={["Footer", "Header"]} />
```
korgon, line 83: Related to question above... would it work like this when using Typescript?

Maybe not important, but we may want to support using lowercase?

korgon (changes requested):
Nice to finally get this tested and a storybook. We should probably look at the hideSidebar button a little closer and look into these enums - see comments.

korgon (changes requested):
Good progress here. I imagine the component evolving in the coming weeks, so this lays a good foundation for improvement.

korgon (approved):

PR #902, opened 2023-09-28
didYouMean Text in SearchHeader

chrisFrazier77:


packages/snap-preact-components/src/components/Atoms/SearchHeader/SearchHeader.tsx
```diff
@@ -104,13 +108,24 @@ export const SearchHeader = observer((properties: SearchHeaderProps): JSX.Elemen
...
+									<h3
+										className={classNames('ss__search-header__title', 'ss__search-header__title--no-results')}
+										aria-atomic="true"
+										aria-live="polite"
+										aria-label={`No results found for ${search?.query?.string}`}
+										dangerouslySetInnerHTML={{ __html: noResultsText as string }}
+									></h3>
+
+									{search?.didYouMean && (
+										<h3
```
korgon, line 121: As more of a sub-title, should this maybe be an h4?

packages/snap-preact-components/src/components/Atoms/SearchHeader/SearchHeader.tsx
```diff
@@ -104,13 +108,24 @@ export const SearchHeader = observer((properties: SearchHeaderProps): JSX.Elemen
...
+											className="ss__search-header__title--no-results__dym"
```
korgon, line 122: All of the others are using the classname `ss__search-header__title` with the modifier class. Perhaps we should keep it consistent with this one as well?

packages/snap-preact-components/src/components/Atoms/SearchHeader/SearchHeader.test.tsx
```diff
@@ -267,6 +267,53 @@ describe('Search Header Component', () => {
...
+
+			const headerElement = rendered.container.querySelector('.ss__search-header__title--no-results__dym');
+			expect(headerElement).toBeInTheDocument();
+			expect(headerElement).toHaveTextContent(customdym);
+		});
+
+		it('renders a custom dymText using a function', async () => {
+			const customdym = (data: any) => {
+				const { pagination, search } = data;
+				return `${pagination.totalResults} ${search?.query?.string} dym!`;
```
korgon, line 295: I realize this test is working perfectly, but maybe the test should be more representative of typical usage?

packages/snap-preact-components/src/components/Atoms/SearchHeader/SearchHeader.tsx
```diff
@@ -139,6 +154,7 @@ export interface SearchHeaderProps extends ComponentProps {
 	subtitleText?: string | ((data: data) => string);
 	correctedQueryText?: string | ((data: data) => string);
 	noResultsText?: string | ((data: data) => string);
+	dymText?: string | ((data: data) => string);
```
korgon, line 157: I realize keeping things short can be helpful to reduce characters and all, but I think we should go with the long variable `didYouMeanText` here to align with the other prop names - eg: `correctedQueryText` (which is even longer).

korgon (changes requested):
Looking good.

korgon (approved):

PR #907, opened 2023-09-29
feat: move until utility function to toolbox

dkonieczek:


packages/snap-toolbox/src/until/until.ts
```diff
@@ -0,0 +1,65 @@
...
+	exponential: number;
+	defer: boolean;
+	executeFunction: boolean;
+};
+
+const defaultUntilOptions: UntilOptions = {
+	checkMax: 20,
+	checkCount: 0,
+	checkTime: 50,
+	exponential: 1.1,
```
dkonieczek, line 14: I lowered this to `1.1` from `1.2`
With the defaults, the last interval with `1.1` is about 300ms compared to about 1900ms when it's `1.2`

packages/snap-toolbox/src/until/until.ts
```diff
@@ -0,0 +1,65 @@
...
+	defer: false,
+	executeFunction: true,
```
dkonieczek, line 16: Added this new `executeFunction` option since by default any function provided will execute before resolving without any function parameters. When this option is `false`, a reference is returned instead.

korgon: I am slightly confused about this change. The main purpose of being able to use a function is for the function to be executed on the check interval and we would wait until said function returns true. This new option does not make sense to me... if you use it, and provided a function for `thing` it would always be truthy (as a function is) and thus would execute as expected.

If you were simply waiting for a variable that is expected to be a function to exist you could write a function to check for that, wait until that is true and then run the function.

packages/snap-toolbox/src/until/until.ts
korgon: Maybe we should change some other defaults to get us around 5-7seconds?
```
1.1: 3150.124972129084
1.2: 11201.279977342421
```

korgon:
```
checkMax: 25,
checkTime: 60,
exponential: 1.1,
```
Yields: 6490.905922636333

packages/snap-toolbox/src/until/until.test.ts
korgon, line 1: Need a test for a variable changing mid run.
Also one that ensures it waits a certain amount of time before it rejects (calculated by options).

korgon (changes requested):
Looking good, couple small requests.

korgon (approved):
Looks good!

PR #909, opened 2023-10-03
Redirect single result

dkonieczek:


packages/snap-controller/src/Autocomplete/AutocompleteController.ts
```diff
@@ -33,6 +33,7 @@ const defaultConfig: AutocompleteControllerConfig = {
 		},
 		redirects: {
 			merchandising: true,
+			singleResult: true,
```
dkonieczek, line 36: on by default?

packages/snap-store-mobx/src/types.ts
```diff
@@ -63,33 +63,36 @@ export type FinderFieldConfig = {
 	levels?: string[];
 };
 
+export type AutocompleteStoreConfigSettings = {
```
dkonieczek, line 66: This was refactored in the templates branch. `singleResult` is the addition here.

korgon (approved):

PR #913, opened 2023-10-06
Radio, RadioList, List, PerPage & SortBy components

chrisFrazier77:


packages/snap-preact-components/src/components/Molecules/RadioSelect/RadioSelect.tsx
```diff
@@ -0,0 +1,174 @@
...
+			},
+
+			'.ss__radio-select__option': {
+				cursor: 'pointer',
+				'& label': {
+					cursor: 'pointer',
+				},
+			},
+
+			'.ss__radio-select--selected': {
```
korgon, line 32: This class doesn't seem to be used anywhere...

packages/snap-preact-components/src/components/Molecules/RadioSelect/RadioSelect.tsx
```diff
@@ -0,0 +1,174 @@
...
+			height: 20,
+			viewBox: undefined,
+			children: <circle cx="10" cy="10" r="5" stroke="black" stroke-width="3" fill="white"></circle>,
+			className: 'ss__list__option__radio__icon--unchecked',
+			// inherited props
+			...defined({
+				disableStyles,
+			}),
+			// component theme overrides
+			theme: props?.theme,
```
korgon, line 88: Do you think we ought to make some radio icons for the library instead of doing it this way?

korgon (changes requested):
After playing around with this a bit more, and considering we will be re-using the same concepts in a "SelectLayout" component in the future, I am a bit concerned with re-using the `RadioSelect` component when choosing the "list" type for `SortBy` and `PerPage`. My main concern has to do with styling. Perhaps a simple "ListSelect" component is a good solution. Let's meet about this soon.

packages/snap-preact-components/src/components/Molecules/RadioList/readme.md
```diff
@@ -11,42 +11,42 @@ Renders a list of options with radio inputs.
...
-<RadioSelect options={store.options} disabled={true} />
+<RadioList options={store.options} disabled={true} />
 ```
 
 ### selected
 The `selected` prop specifies the currently selected Option object. 
 
 ```jsx
-<RadioSelect options={store.options} selected={store.pagination.pageSize} />
+<RadioList options={store.options} selected={store.pagination.pageSize} />
```
korgon, line 49: Similar to `List` component comment - may want to use the `store.pagination.pageSizeOptions` for the options here.

packages/snap-preact-components/src/components/Molecules/SortBy/SortBy.tsx
```diff
@@ -43,7 +44,17 @@ export const SortBy = observer((properties: SortByProps): JSX.Element => {
 		},
 		RadioSelect: {
```
korgon, line 45: RadioList?

packages/snap-preact-components/src/components/Molecules/SortBy/SortBy.tsx
```diff
@@ -111,12 +127,13 @@ export const SortBy = observer((properties: SortByProps): JSX.Element => {
 
 interface SelectSubProps {
 	Select: Partial<SelectProps>;
-	RadioSelect: Partial<RadioSelectProps>;
+	RadioSelect: Partial<RadioListProps>;
```
korgon, line 130: RadioList?

korgon (changes requested):

packages/snap-preact-components/src/components/Molecules/List/List.tsx
```diff
@@ -0,0 +1,139 @@
...
+		<CacheProvider>
+			<div {...styling} className={classnames('ss__list', className)}>
+				{titleText && <h5 className="ss__list__title">{titleText}</h5>}
+
+				<ul className="ss__list__options-wrapper">
+					{options.map((option: option) => {
+						return (
+							<li
+								className={`ss__list__option ${selection == option.value ? 'ss__list__option--selected' : ''} ${
+									disabled ? 'ss__list__option--disabled' : ''
```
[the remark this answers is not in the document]
chrisFrazier77: add disabled prop to individual options && disable the list entirely with other prop

packages/snap-preact-components/src/components/Molecules/Radio/Radio.tsx
korgon, line 1: In testing in Storybook the `size` prop is not working when using the `native` prop... perhaps the size can just be used for width and height in that case?

packages/snap-preact-components/src/components/Molecules/Radio/Radio.tsx
```diff
@@ -0,0 +1,188 @@
...
+			) : (
+				<span
+					{...styling}
+					className={classnames('ss__radio', { 'ss__radio--disabled': disabled }, className)}
+					onClick={(e) => clickFunc(e)}
+					ref={(e) => (!disableA11y ? useA11y(e) : null)}
+					aria-label={`${disabled ? 'disabled' : ''} ${checkedState ? 'checked' : 'unchecked'} checkbox`}
+					role="radio"
+				>
+					{checkedState ? <Icon {...subProps.activeIcon} name="active" /> : <Icon {...subProps.inactiveIcon} name="inactive" />}
```
korgon, line 166: Should these icon names be more specific? Perhaps to match the classname?

packages/snap-preact-components/src/components/Molecules/RadioList/RadioList.tsx
```diff
@@ -116,30 +106,16 @@ export function RadioList(properties: RadioListProps): JSX.Element {
 			<div {...styling} className={classnames('ss__radio-list', className)}>
 				{titleText && <h5 className="ss__radio-list__title">{titleText}</h5>}
 
-				<ul className="ss__radio-list__options-wrapper">
+				<ul className={`ss__radio-list__options-wrapper ${disabled ? 'ss__radio-list__options-wrapper--disabled' : ''}`}>
```
korgon, line 109: Similar to the "List" component, I think it is better to put the `--disabled` modifier class on the root element.

packages/snap-preact-components/src/components/Molecules/RadioList/RadioList.tsx
```diff
@@ -165,10 +141,10 @@ export interface RadioListProps extends ComponentProps {
 type option = {
```
korgon, line 141: To ensure future type compatibility, should all of these components use a shared type instead of redefining it?

packages/snap-preact-components/src/components/Molecules/Radio/Radio.tsx
```diff
@@ -0,0 +1,188 @@
...
+			'&.ss__radio--disabled': {
+				opacity: 0.5,
+			},
+			'& .ss__radio__empty': {
+				display: 'inline-block',
+				width: `calc(${size} - 30%)`,
+				height: `calc(${size} - 30%)`,
+			},
+
+			'& .ss__radio__icon--active circle:first-of-type': {
```
korgon, line 35: These color styles should be unnecessary when making the suggested change to the icon path in this review.

packages/snap-preact-components/src/components/Molecules/Radio/Radio.tsx
```diff
@@ -0,0 +1,188 @@
...
+	const subProps: RadioSubProps = {
+		activeIcon: {
+			// default props
+			className: 'ss__radio__icon--active',
+			icon: checkedIcon || 'bullet',
+			// global theme
+			...globalTheme?.components?.icon,
+			// inherited props
+			...defined({
+				color: color || theme?.colors?.primary,
```
korgon, line 85: I think the `Icon` component already does this fallback, so this may be unnecessary.

packages/snap-preact-components/src/components/Molecules/Radio/Radio.tsx
```diff
@@ -0,0 +1,188 @@
...
+		},
+		inactiveIcon: {
+			// default props
+			className: 'ss__radio__icon--inactive',
+			icon: unCheckedIcon || 'bullet-o',
+			// global theme
+			...globalTheme?.components?.icon,
+			// inherited props
+			...defined({
+				color: color || theme?.colors?.primary,
```
korgon, line 100: I think the `Icon` component already does this fallback, so this may be unnecessary.

packages/snap-preact-components/src/components/Molecules/RadioList/RadioList.test.tsx
```diff
@@ -219,6 +215,7 @@ describe('RadioList Component', () => {
 
 		const rendered = render(<RadioList native={true} options={options} onSelect={selectFn} />);
 
+		rendered.debug();
```
korgon, line 218: Should probably remove.

packages/snap-preact-components/src/components/Molecules/SortBy/readme.md
```diff
@@ -5,6 +5,7 @@ Renders a Select dropdown or a RadioSelect, to be used with the SearchSortingSto
 ## Sub-components
 - Select
 - RadioSelect
```
korgon, line 7: List?

packages/snap-preact-components/src/components/Molecules/Select/Select.stories.tsx
```diff
@@ -81,6 +81,26 @@ export default {
 			},
 			control: { type: 'boolean' },
 		},
+		hideSelection: {
+			description: 'Hide the current selection in the dropdown button',
```
korgon, line 85: Might want to callout "(non-native only)".

packages/snap-preact-components/src/components/Molecules/Select/Select.stories.tsx
```diff
@@ -81,6 +81,26 @@ export default {
...
+			table: {
+				type: {
+					summary: 'boolean',
+				},
+				defaultValue: { summary: false },
+			},
+			control: { type: 'boolean' },
+		},
+		hideIcon: {
+			description: 'Hide the icon in the dropdown button',
```
korgon, line 95: Might want to callout "(non-native only)" - similar to hide`LabelOnSelection` prop.

korgon (changes requested):

packages/snap-preact-components/src/components/Atoms/Icon/Icon.tsx
```diff
@@ -67,7 +67,7 @@ export function Icon(properties: IconProps): JSX.Element {
 			</svg>
 		</CacheProvider>
 	) : (
-		<Fragment></Fragment>
+		<Fragment>h</Fragment>
```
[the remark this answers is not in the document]
chrisFrazier77: I believe you were looking at an older commit here.

packages/snap-preact-components/src/components/Molecules/Radio/Radio.tsx
chrisFrazier77: seemed to be needed still.

packages/snap-preact-components/src/components/Molecules/Radio/Radio.tsx
korgon: This is working now, but there is a pretty significant difference when switching between native and not-native (related to new comment below).

packages/snap-preact-components/src/components/Molecules/List/List.tsx
korgon, line 1: I think the styling could be cleaned up a bit. For example, '.ss__list__option' is the same as 'li', and '.ss__list__options-wrapper' is the same as 'ul' - so why have separate style blocks? Similarly the 'li label' styles could go inside of the '& label' section in '.ss__list__option'.

packages/snap-preact-components/src/components/Molecules/RadioList/RadioList.tsx
korgon, line 1: Similar to `List` comment, I think the styling could be cleaned up here in the same way. For example, '.ss__radio-list__option' is the same as 'li', and '.ss__radio-list__options-wrapper' is the same as 'ul' - so why have separate style blocks? Similarly the 'li label' styles could go inside of the '& label' section in '.ss__radio-list__option'.

korgon (changes requested):
Main concerns are addressed, should be good to merge with review changes.

PR #917, opened 2023-10-10
List Palette Options

chrisFrazier77:
props for layout list or grid, hideCount, & hideCheckbox

### DIFF
--- a/packages/snap-preact-components/src/components/Molecules/FacetPaletteOptions/FacetPaletteOptions.tsx
+++ b/packages/snap-preact-components/src/components/Molecules/FacetPaletteOptions/FacetPaletteOptions.tsx
@@ -22,11 +22,11 @@ const CSS = {
 			flexFlow: 'row wrap',
 			gridTemplateColumns: `repeat(${columns}, calc((100% - (${columns! - 1} * ${gapSize}))/ ${columns}))`,
 			gap: gapSize,
-			padding: '5px 0px',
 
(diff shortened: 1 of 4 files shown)


packages/snap-preact-components/src/components/Molecules/FacetPaletteOptions/FacetPaletteOptions.tsx
korgon, line 1: When using list the checkboxes probably ought to align with the normal list options facets.
image

korgon (changes requested):
Nice work!

packages/snap-preact-components/src/components/Molecules/FacetPaletteOptions/FacetPaletteOptions.tsx
korgon, line 1: Should the `ss__facet-palette-options__option__value` be bold when selected (similar to other list facets)?

korgon (changes requested):
image

When using `layout='list'` the palette swatch sizes grow with the width of the container - this should not be. It might make sense to move to a fixed width with this prop option (perhaps 16px). The flex "gap" is also still in place, causing some spacing issues.

It also seems weird to show the swatch "selected" state and the "X" to remove on hover. I'd suggest that we not show those styles in this case.

image

korgon (commented):
Looks good in the demo! 🙌🏻 

One last small tweak...

korgon (approved):

PR #920, opened 2023-10-13
Filtering in Recommendations

chrisFrazier77:


packages/snap-toolbox/src/charsParams/charsParams.ts
```diff
@@ -15,9 +15,9 @@ export function charsParams(params: ParameterObject): number {
 					return length + keyLength + 1 + ('' + val).length;
```
korgon, line 15: Because `value` here can be an array of anything now (due to new "unknown" type) we will likely get some inaccurate numbers for our count. I think the `'' + val` ought to take care of all potential array types, we may want to add some testing to ensure that we are not throwing any errors when weird types are passed into the function.

packages/snap-toolbox/src/charsParams/charsParams.ts
```diff
@@ -15,9 +15,9 @@ export function charsParams(params: ParameterObject): number {
...
 				}, 0)
 			);
-		} else {
+		} else if (typeof value == 'string' || typeof value == 'number') {
 			return count + keyLength + 1 + ('' + value).length;
-		}
+		} else return count + keyLength;
```
korgon, line 20: Currently with this change objects are allowed to be passed in - but we will not fully traverse them. Instead, we would only get the key length and maybe a length for "[object object]" - which is likely to fall short of the actual number of characters.

It would be nice to make this function recursive to handle any depth of object, as that would get us a more accurate count of the "chars". Additional testing would be required to ensure that a change like that is working as intended.

packages/snap-client/src/Client/apis/Recommend.ts
```diff
@@ -128,7 +146,82 @@ export class RecommendAPI extends API {
...
+				[filter: `filter.${string}`]: string | number | string[] | number[];
+			} = {};
+
+			filters.map((filter) => {
+				if (filter.type == 'value') {
+					//check if filterArray contains a filter for this value already
+					if (filterArray[`filter.${filter.field}`]) {
+						// is the existing filter an array already? or just a single value
+						if (typeof filterArray[`filter.${filter.field}`] == 'string') {
+							filterArray[`filter.${filter.field}`] = [filterArray[`filter.${filter.field}`] as string, filter.value];
```
korgon, line 203: This breaks once the existing value is an array. Better to always have it an array and just push to it.

image

korgon (changes requested):
Great start!

I think it would be beneficial for us to also create a new E2E Recommendation test for filters to ensure it works in the full stack.

packages/snap-toolbox/src/charsParams/charsParams.ts
korgon: Let's move to a separate issue.

chrisFrazier77: ended up being simpler to add than expected.

packages/snap-client/src/Client/transforms/recommendationFiltersPost.test.ts
korgon, line 1: Can you add a wrapping "describe" block similar to other tests?

packages/snap-client/src/types.ts
```diff
@@ -107,6 +107,21 @@ export type RecommendRequestModel = {
...
 	limits?: number | number[];
 	order?: number;
+	filters?: RecommendationRequestFilterModel[];
+};
+
+export type GetRecommendRequestModel = Omit<RecommendRequestModel, 'filters'> & {
+	[filter: `filter.${string}`]: (string | number)[];
+};
+
+export type PostRecommendationRequestFiltersModel = Omit<RecommendRequestModel, 'filters'> & {
```
korgon, line 117: Isn't this just the `PostRecommendationRequestModel`, not sure why it is named differently (see `GetRecommendRequestModel` above).

packages/snap-preact/src/Instantiators/RecommendationInstantiator.tsx
```diff
@@ -192,7 +192,10 @@ export class RecommendationInstantiator {
 				const defaultGlobals = {
 					limits: 20,
 				};
-				const globals = deepmerge(deepmerge(defaultGlobals, this.config.client?.globals || {}), contextGlobals);
+				const globals = deepmerge(
+					deepmerge(deepmerge(defaultGlobals, this.config.client?.globals || {}), contextGlobals),
+					(this.config.config?.globals as any) || {}
```
korgon, line 197: This looks related to how the `options.filters` was incorrectly added in the controller code, instead the "contextGlobals" should be updated above, similarly to how the `options.categories` is done on line 166 above.

packages/snap-toolbox/src/charsParams/charsParams.test.ts
```diff
@@ -38,4 +38,9 @@ describe('param character length', () => {
 		const countWithMixed = charsParams({ array: ['one', 'two'], one: 1, char: 'char' });
 		expect(countWithMixed).toBe(33);
 	});
+
+	it('handles nested objects', () => {
+		const countWithObj = charsParams({ foo: { thing: '4', bar: 'stuff', char: {} } });
+		expect(countWithObj).toBe(28);
```
korgon, line 44: I don't fully understand why this is 28... but it is good that it is closer than it would have been without the recursion.

korgon (changes requested):
We need to get an answer about supporting multiple ranges on the same field for one of my comments in this review, but the rest should be able to be addressed now.

packages/snap-client/src/Client/apis/Recommend.ts
```diff
@@ -114,13 +114,15 @@ export class RecommendAPI extends API {
 
 					// //transform filters here
 					if (batch.request.filters) {
-						(batch.request as PostRecommendRequestModel)['filters'] = this.transformFilters(batch.request.filters, 'POST') as filtersObj[];
+						(batch.request as PostRecommendationRequestFiltersModel)['filters'] = transformRecommendationFiltersPost(
```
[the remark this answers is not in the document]
korgon: Agreeing not to combine because this is likely a user error.

packages/snap-client/src/Client/apis/Recommend.ts
```diff
@@ -106,13 +106,33 @@ export class RecommendAPI extends API {
...
+
+					response = await this.postRecommendations(batch.request as PostRecommendationRequestFiltersModel);
 				} else {
-					response = await this.getRecommendations(batch.request as RecommendRequestModel);
+					if (batch.request.filters) {
+						const filters = transformRecommendationFiltersGet(batch.request.filters);
+						if (filters) {
+							Object.keys(filters).map((filter) => {
+								const _filter = filter as `filter.${string}`;
+								(batch.request as GetRecommendRequestModel)[_filter] = filters[_filter as keyof typeof filters];
```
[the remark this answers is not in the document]
korgon: Agreeing not to combine because this is likely a user error.

packages/snap-preact/src/Instantiators/RecommendationInstantiator.tsx
chrisFrazier77: This was discussed and decided an acceptable bugfix

packages/snap-client/src/Client/apis/Recommend.test.ts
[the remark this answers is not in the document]
korgon: I'd like to see one more test here dealing with the POST version of a similar request.

korgon (changes requested):
Almost there! Just 1 more test please :)

korgon (approved):
Gave this a final test in the demo store and it is working great!

PR #927, opened 2023-10-20
Facet toggle

chrisFrazier77:


packages/snap-preact-components/src/components/Molecules/FacetToggle/FacetToggle.tsx
korgon, line 1: We should split things up a bit more "atomically" here. For example, the toggle part should likely be it's own atom component (likely named `toggle` or `switch`) that would behave similar to the checkbox component. I like the idea of having a `size` prop to control the overall size of the toggle, as well as a `disabled` prop to disable the toggling. I think using the :before pseudo selector is a bit limiting, and I'd rather see a container "wrapper" with a track and switch inside (and maybe the input inside of the switch?).

korgon: The other facet option components typically can take "facet" or "values" as props - I wonder if we can get to a similar interface... I think we would have to assume that the "toggle" display types of facets would have either a one/zero, true/false or yes/no value set.

korgon (changes requested):
This component ought to be broken up a bit more atomically.

The prop interface is a bit awkward, I think we could potentially leverage the facet name as the label unless a different one was provided. The usage of this inside of the `Facet` component could be problematic... I'm pretty sure our API will by default order the options based on the number of products per option; this means that in one search we could pass in "yes" and in another we could pass in "no" which would produce an awkward experience. The best way to possibly handle this might be to require a certain value set for this facet until we get an SMC "toggle" display type.

packages/snap-preact-components/src/components/Molecules/FacetToggle/FacetToggle.tsx
korgon: Any reason we cannot change the interface to be `values` and `facet` and to match the other facet options components?

packages/snap-preact-components/src/components/Atoms/Toggle/Toggle.tsx
```diff
@@ -0,0 +1,145 @@
...
+
+	const clickFunc = (e: React.MouseEvent<Element, MouseEvent>) => {
+		onClick && onClick(e, !toggledState);
+		setToggleState(!toggledState);
+	};
+
+	return (
+		<CacheProvider>
+			<div {...styling} className={classnames('ss__toggle', disabled ? 'ss__toggle--disabled' : '', className)}>
+				{label && <label className="ss__toggle__label">{label}</label>}
```
korgon, line 118: We have two `label` elements but no `input` element... should we keep the two labels? Any reason not to have the input?

packages/snap-preact-components/src/components/Atoms/Toggle/Toggle.tsx
```diff
@@ -0,0 +1,145 @@
...
+				backgroundColor: `${activeColor}`,
+			},
+
+			'& .ss__toggle__switch--filtered .ss__toggle__slider-box .ss__toggle__slider': {
+				transform: `translateX(calc(${size} / 2 - 4px))`,
+			},
+
+			/* Rounded sliders */
+			'& .ss__toggle__slider-box.ss__toggle__slider-box--round': {
+				borderRadius: '34px',
```
korgon, line 74: This specific pixel radius does not scale properly once the `size` goes above 150px... Perhaps some calculation can be done to get the proper radius here?

korgon (changes requested):
Couple things...

packages/snap-preact-components/src/components/Molecules/FacetToggle/readme.md
```diff
@@ -1,17 +1,24 @@
 # Facet Toggle
 
-Renders a facet option toggle switch
+Renders a facet option toggle switch. Note this component will only render when used with a single facet value.
```
korgon, line 3: Maybe rephrase?

```
Note this component will only render when used with a facet that contains a single value.
```

korgon (approved):
Looks good!

PR #933, opened 2023-10-24
Resulttracker

chrisFrazier77:
adding a new tracker for result click beacon tracking

### DIFF
--- a/packages/snap-preact-components/src/components/Organisms/Results/Results.tsx
+++ b/packages/snap-preact-components/src/components/Organisms/Results/Results.tsx
@@ -16,7 +16,7 @@ import { defined, mergeProps } from '../../../utilities';
 import { Theme, useTheme, CacheProvider } from '../../../providers';
 import { useDisplaySettings } from '../../../hooks/useDisplaySettings';
 import { ResultLayout, ResultLayoutTypes } from '../../Layouts/ResultLayout';
-import { ResultTracker } from '../../Trackers/ResultTracker';
+import { SearchResultTracker } from '../../Trackers/SearchResultTracker';
 
 const CSS = {
 	results: ({ columns, gapSize }: Partial<ResultsProps>) =>
@@ -152,7 +152,7 @@ export const Results = observer((properties: ResultsProps): JSX.Element => {
 									return <ResultLayout controller={controller} result={result} layout={resultLayout} />;
 								} else {
 									return (
-										<ResultTracker result={result} controller={controller!}>
+										<SearchResultTracker result={result} controller={controller as SearchController}>
 											<Result
 												key={(result as Product).id}
 												{...subProps.result}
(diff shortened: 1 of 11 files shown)


korgon (changes requested):
I think we ought to rename this component to SearchResultTracker (in a similar naming convention to the recommendation trackers). Additionally, and this can be re-arranged later, the organization in Storybook probably needs to be adjusted to match the recommendation structure.

It may be too late, but this component and the recommendation version are super similar, I wonder if there is a way we can combine them...?

korgon (commented):
Thinking about this a bit more - how will we handle differentiating between search vs. autocomplete in the Results component? Will we need to use a different tracker (eg. SearchResultTracker vs. AutocompleteResultTracker) depending on which controller gets passed in?

packages/snap-preact-components/src/components/Organisms/Results/Results.tsx
```diff
@@ -151,13 +152,15 @@ export const Results = observer((properties: ResultsProps): JSX.Element => {
...
 								} else {
 									return (
-										<Result
-											key={(result as Product).id}
-											{...subProps.result}
-											result={result as Product}
-											layout={props.layout}
-											controller={controller}
-										/>
+										<SearchResultTracker result={result} controller={controller as SearchController}>
```
korgon, line 155: I'm thinking that we should only use the tracker when the controller exists. This way we can ensure that tracking only happens when coupled with a controller. Additionally, this will allow us to "switch" out the tracker (or pass it different props) based on the controller type in the future.

packages/snap-preact-components/src/components/Trackers/SearchResultTracker/SearchResultTracker.tsx
```diff
@@ -49,6 +49,8 @@ export const SearchResultTracker = observer((properties: SearchResultTrackerProp
 		>
 			{children}
 		</div>
+	) : (
+		<Fragment></Fragment>
```
korgon, line 53: I wonder if it makes sense to return the children in the fragment? Consumers may not know it is not setup properly if we did I suppose eh?

korgon (dismissed):

packages/snap-preact-components/src/components/Trackers/SearchResultTracker/SearchResultTracker.tsx
chrisFrazier77: yeah I think your right it probably should return the children. If want to inform of something wrong a console warning is better than breaking and not returning anything

korgon (changes requested):
We decided to render the children even when no controller is provided. When this happens perhaps we should throw an error? This would cause our Cypress tests to fail, which would maybe be more helpful than just logging an error.

korgon (approved):

PR #937, opened 2023-10-31
Update carousel readme to use correct swiper version doc links

chrisFrazier77:


packages/snap-preact-components/src/components/Molecules/Carousel/readme.md
```diff
@@ -1,14 +1,14 @@
 # Carousel
 
-Renders a carousel of slides using children, built with [Swiper](https://swiperjs.com/)
+Renders a carousel of slides using children, built with [Swiper](https://swiper6.vercel.app)
```
korgon, line 3: Maybe we should mention the version of Swiper we are using here?

packages/snap-preact-components/src/components/Molecules/Carousel/readme.md
korgon, line 1: While we are here, maybe edit the `modules` example to have the "required" scrollbar options?

```
<Carousel modules={[Scrollbar]} scrollbar={{ draggable: true }}>{children}</Carousel>
```

korgon (changes requested):
All of this appears to need to be done to the `Recommendation` component docs as well... 😬

korgon (approved):
Looks good.

PR #946, opened 2023-11-14
feat(preact-components-layoutselector): layoutSelector component and …

chrisFrazier77:
…tweaks to list and select


packages/snap-preact-components/src/components/Molecules/LayoutSelector/LayoutSelector.tsx
```diff
@@ -0,0 +1,138 @@
...
+			// global theme
+			...globalTheme?.components?.radioList,
+			// inherited props
+			...defined({
+				disableStyles,
+			}),
+			// component theme overrides
+			theme: props?.theme,
+		},
+		List: {
```
korgon, line 53: Should this add some default props for this component?
Perhaps `multiSelect: false` and/or `horizontal: true` and/or `hideCheckbox: true`?

packages/snap-preact-components/src/components/Molecules/LayoutSelector/LayoutSelector.tsx
```diff
@@ -0,0 +1,138 @@
...
+						onSelect(e, option);
+					}}
+				/>
+			)}
+
+			{type?.toLowerCase() == 'list' && (
+				<List
+					{...styling}
+					className={classnames('ss__layout__list', className)}
+					{...subProps.RadioSelect}
```
korgon, line 97: subProps.List?

packages/snap-preact-components/src/components/Molecules/LayoutSelector/readme.md
```diff
@@ -0,0 +1,91 @@
+# LayoutSelector
+
+Renders a layout selector to be used with snap templates for changing result layout settings. 
+
+## Sub-components
+- Select
+- RadioSelect
```
korgon, line 7: RadioList

packages/snap-preact-components/src/components/Molecules/PerPage/PerPage.tsx
korgon, line 1: Now that the `selected` props are all supposed to be matches to entries in the options I think we could potentially have an issue here when these options are customized.

For example, when using the "select" type, the `selected` prop assumes a specific label of `Show ${store.pageSize}`. Similarly, the "list" and "radio" types assume that there is no label, and only a value.

When testing the "list" type in the demo, none of the options show as selected (I think due to this issue):
image

packages/snap-preact-components/src/components/Molecules/RadioList/RadioList.tsx
```diff
@@ -99,16 +110,27 @@ export function RadioList(properties: RadioListProps): JSX.Element {
...
+								{(option.value as layoutOptionValue)?.icon ? (
+									<Icon
+										{...subProps.Icon}
+										{...(typeof (option.value as layoutOptionValue).icon == 'string'
+											? { icon: (option.value as layoutOptionValue).icon as string }
+											: ((option.value as layoutOptionValue)?.icon as Partial<IconProps>))}
+									/>
+								) : (
+									<label className="ss__radio-list__option__label">{option.label || option.value}</label>
+								)}
```
korgon, line 133: Should we be restricting having both icon and label?

packages/snap-preact-components/src/components/Molecules/Select/Select.tsx
```diff
@@ -253,15 +272,26 @@ export const Select = observer((properties: SelectProps): JSX.Element => {
 								<li
 									ref={(e) => useA11y(e)}
 									role={'link'}
+									aria-disabled={option.di}
```
korgon, line 275: Typo?

packages/snap-preact-components/src/components/Molecules/Select/Select.tsx
korgon, line 1: I'm a bit concerned about such a large change to an existing component, has this been verified/tested as non-breaking?

packages/snap-preact-components/src/components/Molecules/SortBy/SortBy.tsx
korgon, line 1: Same problem here as seen with `PerPage` - the `selected` must now _exactly_ match one of the options passed in - this creates a problem when attempting to customize these options.

packages/snap-preact-components/src/components/Templates/Search/Search.tsx
```diff
@@ -45,11 +43,22 @@ const CSS = {
...
 			'.ss__search__content': {
 				flex: '1 1 0%',
+				padding: '0px 10px',
 			},
 
 			[`@media only screen and (max-width: ${slideOutToggleWidth})`]: {
 				flexDirection: 'column',
 			},
+
+			'.ss__search__content__toolbar--topToolBar': {
```
korgon, line 53: noCamelCaseClassNamesPlease

packages/snap-preact-components/src/components/Templates/Search/Search.tsx
korgon, line 1: I noticed that when I changed the layout with the component near the toolbar, the component in the sidebar did not change to match the selection - shouldn't they be in sync?

As mentioned in a previous conversation, it may make more sense to do all of this layout switching in a shared store (eg. TemplateStore) in the future.

packages/snap-preact-components/src/types.ts
korgon, line 1: I think extending `ListOption` may have been better than creating a new type for this. What is the reason for stuffing the "icon", "columns" and "component" properties into `layoutOptionValue` instead of having them be properties of the `ListOption`? It seems that these are additional configurations, similar to "disabled" that would fit nicely alongside of "value" instead of inside of it.

packages/snap-preact-components/src/components/Templates/Search/Search.tsx
```diff
@@ -329,8 +436,15 @@ export interface SearchProps extends ComponentProps {
...
 	toggleSidebarButtonText?: string;
+	hideLayoutSelector?: boolean;
+	layoutConfig?: layoutConfig;
 }
 
+export type layoutConfig = {
+	options: layoutOption[];
+	default?: layoutOption;
+};
+
```
korgon, line 447: Repeating an option to make it default is repetitive - we should come up with a better way of setting the default.

korgon (changes requested):
As I was testing this the thought came to me that some of these options do not make sense in a "mobile" (or responsive) display - for example when an option is 4 products across. We need to re-think how this should be handled in these situations...

Related is the consideration of how these layouts work with responsive changes that occur in the "Results" component. We still need responsive layouts to occur by default unless a layout is _selected_ (not a default selection, but actual user selection)... and we need to ensure that the stored user selection makes sense in mobile (or ignore it).

Some of these considerations go beyond the task of constructing the component, but are very much involved in how we utilize this component in Snap Templates.

PR #964, opened 2023-12-18
Scroll Restoration Enhancements

korgon:
* adding ability to fire 'restorePosition' event
* adding 'pageshow' window event setting `settings.restorePosition.onPageShow`

### DIFF
--- a/packages/snap-controller/src/Search/SearchController.ts
+++ b/packages/snap-controller/src/Search/SearchController.ts
@@ -134,6 +134,15 @@ export class SearchController extends AbstractController {
 		// restore position
 		if (this.config.settings?.restorePosition?.enabled) {
 			this.eventManager.on('restorePosition', async ({ controller, element }: RestorePositionObj, next: Next) => {
+				// attempt to grab the element from storage if it is not provided
+				if (!element?.selector) {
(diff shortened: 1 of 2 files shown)


packages/snap-controller/src/Search/SearchController.ts
```diff
@@ -134,6 +134,15 @@ export class SearchController extends AbstractController {
 		// restore position
 		if (this.config.settings?.restorePosition?.enabled) {
 			this.eventManager.on('restorePosition', async ({ controller, element }: RestorePositionObj, next: Next) => {
+				// attempt to grab the element from storage if it is not provided
+				if (!element?.selector) {
+					const lastRequest = this.storage.get('lastStringyParams');
+					const storableRequestParams = getStorableRequestParams(JSON.parse(lastRequest));
```
chrisFrazier77, line 140: is there ever a time where lastStringyParams may not exist? should we null check that getter before parsing and passing to getStorableRequestParams?

korgon: Good catch, that is a valid concern. Running `JSON.parse()` on an undefined or even an empty string is an error.

chrisFrazier77 (approved):

PR #967, opened 2024-01-03
fix: bocachica theme polish

dkonieczek:


korgon (approved):
This is a decent amount of polish, but we still need _more_.

We will merge this in, but more polish will be needed later on to make production ready.

PR #979, opened 2024-01-12
AC Redirect Reversion

korgon:
* reverting the default setting for redirect on single result from AC


dkonieczek (changes requested):
default value in the documentation needs updating `packages/snap-controller/src/Autocomplete/README.md`

PR #983, opened 2024-01-18
Add Recommendation Parameters

korgon:
* added support for `products` parameter
* added support for `blockedItems` parameter
* added logic and test to merge `product` and `products` if they are ever used together
* updated documentation to account for changes (using `products` instead of `product` for all examples)


dkonieczek (commented):
Looks good. Should we have a test for `blockedItems`?

dkonieczek (approved):

PR #990, opened 2024-02-05
feat(tracker): add doNotTrack config options to allow certains events…

dkonieczek:
… to be excluded


korgon:
@dkonieczek @chrisFrazier77 
Refactored to move the ssWebPixel to a local storage entry, and checking for that existence within Snap Preact.

korgon (dismissed):
Looks great!

PR #999, opened 2024-02-09
Horizontal Facets

dkonieczek:


korgon (approved):
Still some small tweaks to be made, but the foundation is sound and needs to now be merged into the layout changes to the theme.

PR #1010, opened 2024-03-13
feat(searchresultstore): adding ability to pre select variant options

chrisFrazier77:


korgon (changes requested):
Can we put events into a separate file to clean up constructor (eg. configureSnapFeatures)


We decided on this name/spacing and interface:
```
searchspring.fire('controller/selectVariantOptions', {
  controllers,
  options,
});
```

Lets use explicit exact match or regex only (no string regex):
```
controllers: ['search', /^recommend_/]
```

Needs Readme.

packages/snap-store-mobx/src/Search/Stores/SearchResultStore.test.ts
```diff
@@ -251,6 +251,38 @@ describe('SearchResultStore', () => {
 			});
 		});
 
+		it('can configured to preselect certain variants', () => {
```
korgon, line 254: Does this test actually verify that the preselected values are actually selected...?

packages/snap-store-mobx/src/Search/Stores/SearchResultStore.ts
```diff
@@ -237,6 +237,41 @@ export class Variants {
 					selection.select(firstAvailableOption.value, true);
 				}
 			});
+		} else {
+			// select first available for each selection
```
korgon, line 241: This comment not quite right :|

packages/snap-store-mobx/src/Search/Stores/SearchResultStore.ts
```diff
@@ -237,6 +237,41 @@ export class Variants {
...
+						//see if that option is in the available options
+						const availablePreferedOptions = availableOptions.find((value) => value.value.toLowerCase() == preference.toLowerCase());
+
+						//use it
+						if (availablePreferedOptions) {
+							preferencedOption = availablePreferedOptions;
+						}
+					};
+
+					if (typeof preferedOptions == 'object') {
```
korgon, line 261: Preferred `Array.isArray()` if that is what we are expecting.

korgon: Logic could be simplified if we just always convert to array, or make it require an array.

korgon (changes requested):
Global events will be nice!

packages/snap-preact/src/setupEvents.test.tsx
```diff
@@ -0,0 +1,196 @@
...
+								type: 'product',
+								variants: {
+									makeSelections: makeSelectionsSpy,
+								},
+							},
+							{
+								type: 'banner',
+								variants: {
+									makeSelections: makeSelectionsSpy,
+								},
```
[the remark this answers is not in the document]
chrisFrazier77: shouldnt, but thats why i wanted that in the test, to make sure it handled skipping banner types correctly

packages/snap-preact/README.md
```diff
@@ -252,9 +252,11 @@ const { search } = snap.controllers;
 ```
 
 ## properties
-
 After instantiating an instance of Snap, the following properties can be accessed. 
 
+### event-manager
```
korgon, line 257: I think the property name should match up with the subheading here - `eventManager`.

packages/snap-store-mobx/src/Search/Stores/SearchResultStore.ts
```diff
@@ -181,17 +181,17 @@ export class Variants {
...
 			this.active = variant;
 			mask.set({ mappings: this.active.mappings, attributes: this.active.attributes });
 		};
 
-		this.update(variantData, preselectedOptions);
+		this.update(variantData, config);
 	}
 
-	public update(variantData: VariantData[], preselectedOptions?: Record<string, string[]>) {
+	public update(variantData: VariantData[], config?: VariantConfig) {
```
korgon, line 194: If this function is called externally (in middleware for example), will the user need to pass the `config` parameter used at construction to get the same settings?

korgon (approved):

PR #1011, opened 2024-03-15
Reactivity Fix

korgon:
* changed `result.display` from updating / merging display, to using a getter for display
* `result.display` computed change resulted in addition of a "mask" to create display
* fixed bug in variant selection, selection will now use `previouslySelected` value
* added tests for mask, display and variants code
* added E2E test for middleware to help prevent reactivity regressions when using `result.display`


dkonieczek:
Looks like `previouslySelected` functionality isn't working. To be further tested on Monday

korgon:
Previously selected functionality cleared up.

chrisFrazier77 (approved):

dkonieczek (approved):

PR #1017, opened 2024-03-28
setVariants function

chrisFrazier77:


packages/snap-store-mobx/src/Search/Stores/SearchResultStore.ts
korgon, line 1: Alternatively we could put this in `variants.update()`.

This does however require that variants have been setup in the config, and restricts fully frontend built variants (perhaps unnecessarily).

korgon: ^ let's go with this

packages/snap-store-mobx/src/Search/Stores/SearchResultStore.ts
```diff
@@ -228,7 +263,7 @@ export class Variants {
 		}
 	}
 
-	public update(fromSelection: VariantSelection) {
+	public refineOptions(fromSelection: VariantSelection) {
```
korgon, line 266: `refineSelections` and rename same named function below (inside VariantSelection) to `refineValues`.

korgon (changes requested):

korgon (approved):

PR #1022, opened 2024-04-04
feat(snap-platforms)

chrisFrazier77:
snap-platforms added for shopify, bigcommerce, & magento2


packages/snap-platforms/package.json
```diff
@@ -0,0 +1,47 @@
...
+	"repository": "https://github.com/searchspring/snap",
+	"publishConfig": {
+		"access": "public"
+	},
+	"scripts": {
+		"build": "rm -rf ./dist && rm -fr ./components/dist && tsc && tsc -p tsconfig.cjs.json",
+		"build:docs": "typedoc --out docs src/index.ts",
+		"dev": "tsc --watch",
+		"format": "prettier --write 'src/**/*.{js,jsx,ts,tsx}'",
+		"lint": "eslint '{src,components}/**/*.{js,jsx,ts,tsx}'",
```
korgon, line 16: "lint": "eslint './**/*.{js,jsx,ts,tsx}'",

korgon (changes requested):
This is the way.

Need to clean up the readme files and error logs to remove plugin references.

packages/snap-platforms/bigCommerce/README.md
korgon, line 1: Good changes!

korgon (commented):
This is looking good. I'd like to discuss some more about the "callback" and async usage of these `addToCart` functions before we set the interface/usage in stone.

packages/snap-platforms/bigCommerce/src/addToCart.ts
```diff
@@ -0,0 +1,110 @@
...
+			if (response.status === endpoint.success) {
+				const jsonResponse = response.json();
+				return jsonResponse;
+			} else {
+				return new Error(`Error: Snap-platform-bigcommerce addToCart responded with ${response.status}, ${response}`);
+			}
+		} catch (err) {
+			console.error(err);
+		}
+	};
```
korgon, line 95: Should this be a separate function outside of this function?

packages/snap-platforms/bigCommerce/src/addToCart.ts
```diff
@@ -0,0 +1,110 @@
...
+					level = level[field];
+				} else {
+					console.error('Error: couldnt find column in item data. please check your idFieldName is correct in the config.');
+					return false;
+				}
+			});
+			if (level && level !== item) {
+				sku = level;
+			} else {
+				sku = item.display.mappings.core?.uid;
```
korgon, line 40: If it is an Id, we should probably call it that (even internally).

korgon (commented):
Types
* add `available` in ListOption type
* refactor List, Grid, Swatches to new type - add
* need class modifiers in components based on ListOption value `--disabled` `--unavailable`
* styling should be the same for class modifiers above
* disabled options should not be clickable
```
export type ListOption = {
	value: string | number;
	label?: string;
	disabled?: boolean;
	available?: boolean;
	default?: boolean;
	icon?: string | Partial<IconProps>;
	url?: UrlManager;
};
```

Grid Component
* List component has a 'title' why not this one?
* need story for overflowing grid and for options with backgrounds (image or whatever)
* prop for hiding "show less" (when you don't want to show less)
* :before pseudo element for "disabled" strike should not have fixed width

List Component
* remove `clickableDisabledOptions`
* instead of padding lets use `gap` from flex
* `disabled` still allows for selections
* when using "horizontal" we should put the title `<h5>` next to `<ul>` by adding styles to main wrapper: `display: flex;align-items: center;`

Swatches Component
* has a fixed width of 40px per slide, why?
* :before pseudo element for "disabled" strike should not have fixed width
* :before pseudo element doesn't work on white swatch
* disabled styling differs from grid / carousel - maybe unify?

VariantSelection Component
* storybook type selection doesn't work
* dropdown type needs `position: relative;` for disabled strike
* for disabled class, maybe use simpler 'line-through' for styling?

korgon (commented):
Could use some final review and testing.

korgon (approved):

korgon (approved):
Looks good to me.

PR #1031, opened 2024-04-22
bugfix for carousel classnames

chrisFrazier77:
add backwards compatible classnames on each rerender of carousel


korgon (approved):
Mhmmm.

PR #1034, opened 2024-04-25
Variant selection components

chrisFrazier77:


packages/snap-preact-components/src/components/Molecules/Swatches/Swatches.stories.tsx
korgon, line 1: **Idea:** This component looks and works great. I am wondering though if we should split this up into two different components:

1. Swatches
2. SwatchesCarousel (this component)

`Swatches` would simply be this component without the carousel with various props for limiting the number of swatches, choosing a display style (eg. box vs. circle, or whatever) and anything else we want to throw in there. This would likely replace "grid" as it has similar function but is more tailored to swatches in general. One idea for this would be showing a "+x", where x is the number of remaining swatches that have been limited. This component wouldn't work very well for bundled recommendations, but it is more of a general purpose swatches component.

`SwatchesCarousel` would be this component, but it would consume `Swatches` (but wouldn't have a limit normally).

packages/snap-preact-components/src/components/Molecules/VariantSelection/VariantSelection.tsx
```diff
@@ -0,0 +1,212 @@
...
+		...globalTheme?.components?.variantSelection,
+		// props
+		...properties,
+		...properties.theme?.components?.variantSelection,
+	};
+
+	const { type, selection, hideTitle, disableStyles, className, style } = props;
+
+	const subProps: VariantSelectionSubProps = {
+		select: {
```
korgon, line 35: This component doesn't seem to handle "disabled" display like the others do, but is it possible? If not, is that a deal breaker for this being an option here?

packages/snap-preact-components/src/components/Molecules/VariantSelection/VariantSelection.tsx
```diff
@@ -0,0 +1,212 @@
...
+						case 'select':
+							return (
+								<Fragment>
+									{(() => {
+										const convertedValues: Option[] = selection.values.map(
+											(val) =>
+												(val = {
+													//@ts-ignore - disabled isnt available on SelectionValue
+													disabled: !val.available,
+													label: val.label || val.value,
```
korgon, line 105: Is label ever not defined on a selection option value?

packages/snap-preact-components/src/components/Molecules/VariantSelection/VariantSelection.tsx
```diff
@@ -0,0 +1,212 @@
...
+							return (
+								<Fragment>
+									{(() => {
+										const convertedValues: ListOption[] = selection.values.map(
+											(val) =>
+												(val = {
+													//@ts-ignore - disabled isnt available on SelectionValue
+													disabled: !val.available,
+													...val,
+												})
```
korgon, line 126: This is repeated a lot, can we just define this once?

Or... to remove it entirely should we just add `disabled` to SelectionValue?

packages/snap-preact-components/src/components/Molecules/VariantSelection/VariantSelection.tsx
```diff
@@ -0,0 +1,212 @@
...
+interface VariantSelectionSubProps {
+	select: SelectProps;
+	list: ListProps;
+	grid: GridProps;
+	swatches: SwatchesProps;
+}
+
+export interface VariantSelectionProps extends ComponentProps {
+	selection: VariantSelectionType;
+	type?: 'select' | 'swatches' | 'grid' | 'list';
```
korgon, line 210: With my proposed changes above, this would become:

```
type?: 'select' | 'swatches' | 'carousel' | 'list';
```

korgon (changes requested):
Initial review, this is looking good, a few comments and ideas to ponder...

korgon:
grid component prop ideas:

```
disableOverflow
disableOverflowAction
overflowButton
overflowButtonInGrid
onOverflowButtonClick
```

PR #1046, opened 2024-05-21
Fix Build Issues

korgon:
* allows for rebuilding of `package-lock.json` again!
* fixing build error due to dependency versioning


dkonieczek (approved):
Tested out locally, looks good!

PR #1061, opened 2024-05-31
variant availability bugfix

chrisFrazier77:
needed to filter out unavailable options even if there are no selections

### DIFF
--- a/packages/snap-store-mobx/src/Search/Stores/SearchResultStore.ts
+++ b/packages/snap-store-mobx/src/Search/Stores/SearchResultStore.ts
@@ -511,13 +511,6 @@ export class VariantSelection {
 						),
 					};
 
-					console.log(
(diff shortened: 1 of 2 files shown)


packages/snap-store-mobx/src/Search/Stores/SearchResultStore.ts
```diff
@@ -478,10 +478,14 @@ export class VariantSelection {
 		let availableVariants = variants.data;
```
korgon, line 478: Any reason not to just set this to filter out all available variants from the start?

```
let availableVariants = variants.data.filter((variant) => variant.available);
```

Then the if statement is unnecessary and we can remove `&& variant.available` from the for loop filter?

korgon: Seems a bit odd that we were not already doing this... 🤔

dkonieczek (approved):

PR #1077, opened 2024-06-20
Pagination Page Size Options 

chrisFrazier77:
filter out pagination page size options over api limit of 100


korgon (approved):
Filtering out the page size options over 100 is a good solution 👍🏻

PR #1078, opened 2024-06-20
feat(tracker): add currency context, add transactionTotal to order tr…

dkonieczek:
…ansaction payload

### DIFF
--- a/packages/snap-tracker/src/Tracker.ts
+++ b/packages/snap-tracker/src/Tracker.ts
@@ -89,6 +89,12 @@ export class Tracker {
 			},
 		};
 
+		if (this.globals.currency) {
+			this.context.currency = {
+				code: this.globals.currency,
+			};
+		}
+
 		if (!window.searchspring?.tracker) {
 			window.searchspring = window.searchspring || {};
 			window.searchspring.tracker = this;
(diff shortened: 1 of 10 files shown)


chrisFrazier77 (dismissed):

packages/snap-tracker/src/Tracker.ts
```diff
@@ -89,6 +89,12 @@ export class Tracker {
 			},
 		};
 
+		if (this.globals.currency) {
```
korgon, line 92: `this.globals.currency` should be of type:
```
{
  code: string;
}
```

packages/snap-tracker/src/Tracker.ts
```diff
@@ -546,6 +571,15 @@ export class Tracker {
 		}
 	};
 
+	setCurrency = (currency: string): void => {
```
korgon, line 574: This function should take an object as the param:
```
{
  code: string;
}
```

packages/snap-tracker/src/Tracker.ts
```diff
@@ -418,7 +424,7 @@ export class Tracker {
 			},
 		},
 		cart: {
-			view: (data: CartViewEvent, siteId?: string): BeaconEvent | undefined => {
+			view: (data: CartViewEvent, siteId?: string, currency?: string): BeaconEvent | undefined => {
```
korgon, line 427: I think we can remove the currency param here and instead assume that the currency has already been set on the context.

packages/snap-tracker/src/Tracker.ts
```diff
@@ -475,10 +490,10 @@ export class Tracker {
 			},
 		},
 		order: {
-			transaction: (data: OrderTransactionData, siteId?: string): BeaconEvent | undefined => {
+			transaction: (data: OrderTransactionData, siteId?: string, currency?: string): BeaconEvent | undefined => {
```
korgon, line 493: I think we can remove the currency param here and instead assume that the currency has already been set on the context.

korgon (changes requested):
A few changes requested in review, but additionally, there should be a way to set the currency via the [script context variables](https://searchspring.github.io/snap/#/integration-context) (similar to `shopperId`).

Perhaps something like this:
```
<script src="https://snapui.searchspring.io/[your_site_id]/bundle.js">
    currency = { code: "USD" };
</script>
```

korgon (approved):
Looks good.

PR #1080, opened 2024-06-20
wrap price symbol in a label for styling

chrisFrazier77:


packages/snap-preact-components/src/components/Atoms/Price/Price.tsx
```diff
@@ -76,7 +75,9 @@ export function Price(properties: PriceProps): JSX.Element {
 	) : (
 		<CacheProvider>
 			<span {...styling} className={classnames('ss__price', { 'ss__price--strike': lineThrough }, className)}>
+				{symbol && !symbolAfter ? <label className={'ss__price__symbol'}>{symbol}</label> : <></>}
 				{formattedPrice}
```
korgon, line 79: We should wrap this in a span with classname `ss__price__value` (or similar).

packages/snap-preact-components/src/components/Atoms/Price/Price.tsx
```diff
@@ -76,7 +75,9 @@ export function Price(properties: PriceProps): JSX.Element {
 	) : (
 		<CacheProvider>
 			<span {...styling} className={classnames('ss__price', { 'ss__price--strike': lineThrough }, className)}>
+				{symbol && !symbolAfter ? <label className={'ss__price__symbol'}>{symbol}</label> : <></>}
```
korgon, line 78: Let's use span instead of label for the symbol.

korgon (changes requested):

korgon (approved):
Looks good!

PR #1084, opened 2024-06-25
call updateSlideClasses onResize to update disabled classnames

chrisFrazier77:


korgon (approved):
Does this also update the aria attribute?

PR #1086, opened 2024-06-27
dont pass breakpoints to swiper

chrisFrazier77:


korgon (approved):
Nice one-liner... seems scary, but I think this ought to be fine, as we are already passing on the props to Swiper via our own breakpoints.

PR #1088, opened 2024-06-27
Preact template themes

dkonieczek:


korgon (approved):
Good fixes here in addition to allowing for themes in Snap Preact.

PR #1105, opened 2024-07-11
build: merging demo stores & webpack refactor

dkonieczek:


packages/snap-preact/components/src/components/Molecules/Result/Result.test.tsx
```diff
@@ -92,6 +92,7 @@ describe('Result Component', () => {
 
 	it('should display a fallback image', () => {
 		mockResults[1].mappings!.core!.imageUrl = '';
+		(mockResults[1] as Product).images = undefined;
```
korgon, line 95: We should revisit this, as I'm pretty sure we don't want to use `images` here any more, instead we will favor the use of the mask.

korgon: (need to remove it from the store)

packages/snap-preact/components/tests/cypress/component/molecules/variantSelection.cy.tsx
```diff
@@ -45,13 +45,13 @@ const controller = new RecommendationController(recommendConfig, {
...
 });
 
-const theme = {
-	components: {
-		variantSelection: {
-			className: 'classy',
-		},
-	},
-};
+// const theme = {
```
korgon, line 48: I'm guessing this was unused?

korgon (approved):

PR #1109, opened 2024-07-11
bugfix for cartstore handling floating 0s for pricing

chrisFrazier77:


korgon (approved):
Strange maths, but works!

PR #1121, opened 2024-07-30
Named theme components

dkonieczek:


korgon:
Let's work to get tests passing in addition to adding more tests to ensure coverage of this core functionality.

korgon (changes requested):

PR #1122, opened 2024-07-30
fix(preact-components-recommendationbundle): adding hideSeed prop

chrisFrazier77:


packages/snap-preact-components/tests/cypress/component/molecules/bundleRecommendation.cy.tsx
```diff
@@ -337,6 +337,48 @@ describe('RecommendationBundle Component', async () => {
...
+		const carouselProps = {
+			enabled: false,
+		};
+		mount(<RecommendationBundle controller={controller} hideSeed={true} carousel={carouselProps} onAddToCart={cy.stub().as('onAddToCart')} />);
+
+		cy.get('.ss__recommendation-bundle').should('exist');
+		cy.get('.ss__recommendation-bundle .ss__recommendation-bundle__wrapper__selector--seed').should('not.exist');
+
+		cy.get('.ss__recommendation-bundle__wrapper').should('exist');
+		cy.get('.ss__recommendation-bundle__wrapper__selector').should('have.length', controller.store.results.length - 1);
```
korgon, line 379: I assume this is 4? Perhaps we should use a real value here since the test above does the same? Or we should change the test above to use the store.results.length?

packages/snap-preact-components/src/components/Trackers/Recommendation/ResultTracker/RecommendationResultTracker.tsx
korgon, line 1: We ought to add this new `track` prop (and its various configuration) to the test in `RecommendationResultTracker.test.tsx`.

packages/snap-preact-components/src/components/Organisms/RecommendationBundle/RecommendationBundle.tsx
korgon, line 1: When using `hideSeed`, as expected, it does not render, and so the `render` product beacon event does not happen for the seed product - nice!

korgon (changes requested):
This is working good. Please see comments regarding testing changes.

packages/snap-preact-components/src/components/Trackers/Recommendation/ResultTracker/RecommendationResultTracker.tsx
korgon: If this is not possible, then perhaps we will need to add it to the E2E Cypress tests in the demo store.

packages/snap-preact-components/src/components/Trackers/Recommendation/ResultTracker/RecommendationResultTracker.test.tsx
```diff
@@ -182,6 +188,240 @@ describe('RecommendationResultTracker Component', () => {
...
+			});
+		});
+
+		trackfn.mockClear();
+
+		const resultElem = rendered.container.querySelector('.findMe');
+
+		await userEvent.click(resultElem!);
+
+		expect(trackfn).toHaveBeenCalledWith({
```
korgon, line 401: It seems the test is verifying that the profile click event has happened, but how can we be sure that the `trackfn` _has not_ been called with a product click event?

korgon (commented):
Thanks for adding additional testing.

korgon (approved):

PR #1125, opened 2024-08-02
Lang props

chrisFrazier77:


packages/snap-preact/components/src/components/Organisms/Facet/Facet.tsx
```diff
@@ -393,15 +417,11 @@ const FacetContent = (props: any) => {
...
-									{...subProps.showMoreLessIcon}
-									{...(((facet as ValueFacet).overflow?.remaining || 0) > 0
-										? { ...(typeof iconOverflowMore == 'string' ? { icon: iconOverflowMore } : (iconOverflowMore as Partial<IconProps>)) }
-										: { ...(typeof iconOverflowLess == 'string' ? { icon: iconOverflowLess } : (iconOverflowLess as Partial<IconProps>)) })}
-								/>
-							}
-							<span>{((facet as ValueFacet)?.overflow?.remaining || 0) > 0 ? showMoreText : showLessText}</span>
+							<Icon
+								{...subProps.showMoreLessIcon}
+								icon={((facet as ValueFacet).overflow?.remaining || 0) > 0 ? iconOverflowMore : iconOverflowLess}
```
korgon, line 422: Change removes support for using IconProps in `iconOverflowMore` and `iconOverflowLess`.

packages/snap-preact/components/src/components/Organisms/Facet/Facet.tsx
```diff
@@ -307,9 +334,7 @@ export const Facet = observer((properties: FacetProps): JSX.Element => {
 							ref={(e) => useA11y(e, disableCollapse ? -1 : 0)}
 							role="heading"
 							aria-level={3}
-							aria-label={`currently ${facet?.collapsed ? 'collapsed' : 'open'} ${facet.label} facet dropdown ${
-								(facet as ValueFacet).values?.length ? (facet as ValueFacet).values?.length + ' options' : ''
-							}`}
+							{...mergedLang.dropdownButton}
 						>
 							{facet?.label}
 							{!disableCollapse && (
```
korgon, line 340: This dropdown icon would be replaced when using the `dropdownButton` value lang prop. Proposal to split the attributes and value in the `useLang` response.

packages/snap-preact/components/src/hooks/useLang.tsx
korgon, line 1: To refactor as described in main PR comment.

packages/snap-preact/components/src/hooks/useLang.tsx
```diff
@@ -0,0 +1,83 @@
...
+	dangerouslySetInnerHTML?: { __html: string };
+	'aria-label'?: string;
+	'aria-valuetext'?: string;
+	title?: string;
+	alt?: string;
+};
+
+export type LangType<P> = string | ((data?: P) => string);
+
+export interface langObjType {
```
korgon, line 16: Capitalize please :)

`LangObjType`

korgon (changes requested):
There are some components for which the a11y attributes are not aligning with the focusable elements (for screen reading) which will cause loss of accessibility function. In addition, some components with this situation have the 'value' of the lang prop names in a different element than the attributes. This necessitates some clever workarounds - preferably splitting the response of `useLang` to return both `attributes` and `value` so that they can be spread onto different elements.

For lang props options that don't support `value` or `attributes` we could create separate types to differentiate them, eg: `Lang`, `LangValue` and `LangAttributes`.

korgon (changes requested):
Almost there! A few changes to push up after pair review.

PR #1126, opened 2024-08-05
Testing CLS

korgon:
* changes for testing CLS


korgon:
Not a good approach for improving CLS.

PR #1132, opened 2024-08-14
feat: add onTarget to template recommendations config

dkonieczek:


packages/snap-preact/src/Templates/Stores/LibraryStore.ts
```diff
@@ -80,20 +82,25 @@ export class LibraryStore {
...
 				},
-				RecommendationBundle: async () => {
+			},
+			'recommendation.default': {
+				Recommendation: async () => {
 					return (
-						this.components.recommendation.RecommendationBundle ||
-						(this.components.recommendation.RecommendationBundle = (await import('./library/components/RecommendationBundle')).RecommendationBundle)
+						this.components['recommendation.default'].Recommendation ||
+						(this.components['recommendation.default'].Recommendation = (await import('./library/components/Recommendation')).Recommendation)
```
korgon, line 99: Is there any reason to not actually go another level deeper in the object instead of using a `.` to delimit the path superficially?

korgon (commented):
Should we go deeper with the object? or no?

packages/snap-preact/src/Templates/Stores/LibraryStore.ts
dkonieczek: No particular reason. I think I'd actually prefer a `/` instead of `.` since we don't do this else where. Nested object would also work

korgon: Unless there is good reason, I think my preference here is nested object.

PR #1133, opened 2024-08-15
recommendationList 

chrisFrazier77:
adding new recommendationList component


packages/snap-preact/components/src/components/Organisms/RecommendationList/RecommendationList.tsx
```diff
@@ -0,0 +1,146 @@
...
+import { RecommendationResultTracker } from '../../Trackers/Recommendation/ResultTracker';
+
+const CSS = {
+	results: ({ columns, gapSize }: Partial<RecommendationListProps>) =>
+		css({
+			display: 'flex',
+			flexFlow: 'row wrap',
+			gap: gapSize,
+			gridTemplateRows: 'auto',
+			gridTemplateColumns: `repeat(${columns}, 1fr)`,
```
korgon, line 23: Shouldn't these go inside of the `'@supports (display: grid)'` section?

packages/snap-preact/components/src/components/Organisms/RecommendationList/RecommendationList.tsx
```diff
@@ -0,0 +1,146 @@
...
+				boxSizing: 'border-box',
+				flex: '0 1 auto',
+				width: `calc(${100 / columns!}% - (${columns! - 1} * ${gapSize} / ${columns} ) )`,
+				marginRight: gapSize,
+				marginBottom: gapSize,
+
+				[`&:nth-of-type(${columns}n)`]: {
+					marginRight: '0',
+				},
+			},
```
korgon, line 35: These do not appear to be necessary.

packages/snap-preact/components/src/components/Organisms/RecommendationList/RecommendationList.tsx
```diff
@@ -0,0 +1,146 @@
...
+			'@supports (display: grid)': {
+				display: 'grid',
+
+				'& .ss__result': {
+					width: 'initial',
+					flex: undefined,
+					margin: 0,
+				},
```
korgon, line 43: These do not appear to be necessary.

packages/snap-preact/components/src/components/Organisms/RecommendationList/RecommendationList.tsx
```diff
@@ -0,0 +1,146 @@
...
+								return <ResultComponent controller={controller} result={result as Product} theme={theme} />;
+							} else {
+								return (
+									<RecommendationResultTracker result={result as Product} controller={controller}>
+										<Result key={(result as Product).id} {...subProps.result} result={result as Product} controller={controller} />
+									</RecommendationResultTracker>
+								);
+							}
+						})()
+					)}
```
korgon, line 124: Should wrap these in `<div className="ss__recommendation-list__results">` and change the styles above to target this element so that the title is not included in the grid.

packages/snap-preact/components/src/components/Organisms/RecommendationList/index.ts
korgon, line 1: Need to also export the component from the library so that it can be properly imported externally (within `snap-preact/components/src/index.ts`).

korgon (changes requested):
Should we rename this to `RecommendationGrid` ?

There are some weird typings going on that prevent me from using the <Result/> component manually as a prop within `resultComponent` here - seems like a problem in all of the Recommendation components as well as the Results component. Can you please correct this?
```
<RecommendationList controller={controller} title={'Recommended For You'} resultComponent={<Result/>}></RecommendationList>
```

It might be nice to have the columns / rows be more dynamic ... by that I mean that if you specify `columns={1}` and don't specify any `rows` prop value, it would be nice to have it render all the results without having to specify the number of `rows`. Similarly, I cannot currently specify `rows={2}` (without also specifying `columns`) - would be nice to have it automatically split the results into two rows; currently doing this only renders a single row. If we could make this more dynamic I think it would be great.

korgon (commented):
Once we merge `develop` into `templates` we will need to add `lazyRender` to this component.

packages/snap-preact/components/src/components/Organisms/RecommendationGrid/RecommendationGrid.stories.tsx
```diff
@@ -134,11 +141,11 @@ Grid.loaders = [
 	},
 ];
 
-export const List = (args: RecommendationListProps, { loaded: { controller } }: { loaded: { controller: RecommendationController } }) => {
-	return <RecommendationList {...args} controller={controller} results={controller?.store?.results} rows={1} />;
+export const Grid = (args: RecommendationGridProps, { loaded: { controller } }: { loaded: { controller: RecommendationController } }) => {
+	return <RecommendationGrid columns={4} {...args} controller={controller} results={controller?.store?.results} />;
```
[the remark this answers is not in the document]
korgon: @chrisFrazier77 templates branch has been updated and now includes `lazyRender` prop.

korgon (approved):
Looks good!

PR #1134, opened 2024-08-19
lazyRender prop

chrisFrazier77:


dkonieczek (approved):

korgon (approved):
This is a step in the right direction. I'm wondering if 10% is the right default value though... I'll verify these changes with the implementations leadership prior to merging.

PR #1135, opened 2024-08-19
esLint improvements

chrisFrazier77:


korgon (commented):
:confused: ... why did these assertions need to be changed?

PR #1142, opened 2024-08-28
Develop -> Templates

korgon:
Welp, this was a bit more effort than expected.


korgon:
Didn't think it had been 3 months since we last did this... 🤔

PR #1153, opened 2024-09-03
refactor: template store constructor interface typing

dkonieczek:


packages/snap-preact/src/Templates/Stores/TemplateStore.ts
korgon, line 1: What about the constructor in the TemplateStore?

korgon: Similar thing there with config vs. settings - a bit confusing I think to have both...

korgon (changes requested):

PR #1154, opened 2024-09-03
refactor: move attribute tracking from Tracker to Snap

dkonieczek:


packages/snap-controller/src/Recommendation/README.md
```diff
@@ -8,7 +8,7 @@ The `RecommendationController` is used when making queries to the API `recommend
 |---|---|:---:|:---:|
 | id | unique identifier for this controller | ➖ | ✔️ |
 | tag | unique name of the recommendations profile | ➖ | ✔️ |
-| realtime | update recommendations if cart contents change (requires [cart attribute tracking](https://github.com/searchspring/snap/blob/main/docs/INTEGRATION.md)) | ➖ |   |
+| realtime | update recommendations if cart contents change (requires [cart attribute tracking](https://github.com/searchspring/snap/blob/main/docs/INTEGRATION_TRACKING.md)) | ➖ |   |
```
korgon, line 11: Good catch!

packages/snap-preact/src/Snap.tsx
```diff
@@ -265,6 +265,65 @@ export class Snap {
 	};
 
 	public handlers = {
```
korgon, line 267: Maybe we could put the different handlers into a separate file (similar to the events) to keep things tidy?

packages/snap-preact/src/setupEvents.test.tsx
```diff
@@ -193,4 +207,67 @@ describe('setupEvents', () => {
...
+	describe('controller/updateRecs', () => {
+		it('can listen for controller/updateRecs event and invoked without parameters', () => {
+			expect(makeSearchSpy).not.toHaveBeenCalled();
+
+			//no data passed at all
+			expect(() => {
+				eventManager.fire('controller/updateRecs');
+			}).not.toThrow();
+
+			expect(makeSearchSpy).toHaveBeenCalledTimes(1);
```
korgon, line 220: We should test that multiple controllers search function was called (would need multiple spies perhaps?).

packages/snap-preact/src/setupEvents.test.tsx
```diff
@@ -193,4 +207,67 @@ describe('setupEvents', () => {
...
+		});
+
+		it('can pass exact match controllerId', () => {
+			expect(makeSearchSpy).not.toHaveBeenCalled();
+			const controllerIds = 'recommend_similar_0';
+			expect(() => {
+				eventManager.fire('controller/updateRecs', { controllerIds });
+			}).not.toThrow();
+
+			expect(makeSearchSpy).toHaveBeenCalledTimes(1);
```
korgon, line 232: How can we be sure the search was called on the controller we provide the `id` for?

packages/snap-preact/src/setupEvents.test.tsx
```diff
@@ -193,4 +207,67 @@ describe('setupEvents', () => {
...
+		});
+
+		it('can pass partial regex match controllerId', () => {
+			expect(makeSearchSpy).not.toHaveBeenCalled();
+			const controllerIds = [/^recommend_/];
+			expect(() => {
+				eventManager.fire('controller/updateRecs', { controllerIds });
+			}).not.toThrow();
+
+			expect(makeSearchSpy).toHaveBeenCalledTimes(1);
```
korgon, line 244: Again test should verify the right controller search function was called.

packages/snap-preact/src/setupEvents.test.tsx
```diff
@@ -193,4 +207,67 @@ describe('setupEvents', () => {
...
+		});
+
+		it('only invokes for recommendation type and if exists', () => {
+			expect(makeSearchSpy).not.toHaveBeenCalled();
+			const controllerIds = ['recommend_similar_0', 'search', 'dne'];
+			expect(() => {
+				eventManager.fire('controller/updateRecs', { controllerIds });
+			}).not.toThrow();
+
+			expect(makeSearchSpy).toHaveBeenCalledTimes(1);
```
korgon, line 256: Again test should verify the right controller search functions were called for provided `id` values.

packages/snap-preact/src/setupEvents.test.tsx
```diff
@@ -43,6 +45,11 @@ describe('setupEvents', () => {
 					},
 				},
 				recommend_similar_0: {
```
korgon, line 47: I think we need a third controller to verify completely.

packages/snap-preact/src/setupEvents.ts
korgon, line 1: We can probably re-use the controller id matching bit if we make that a function...

packages/snap-preact/src/setupEvents.ts
```diff
@@ -48,5 +48,40 @@ export const setupEvents = () => {
 		await next();
 	});
 
+	type ControllerUpdateRecsData = {
+		controllerIds?: (string | RegExp)[];
+	};
+
+	eventManager.on('controller/updateRecs', async (data: ControllerUpdateRecsData, next: Next) => {
```
korgon, line 55: I wonder if we should make this more generic ... maybe `controller/search` or `controller/update` - `update` doesn't sound like what it is actually doing, I'm not sure which is better... 

Ideas:
```
eventManager.on('controller/updateRecs');  // current
eventManager.on('controller/search');  // not restrictive enough, but generic for future usage
eventManager.on('controller/searchRecs');  // similar to search but with additional filtering
eventManager.on('controller/recommend/search');  // similar to search but with additional filtering
```

Just thoughts, I'd like to have a good structure and naming convention for these global events.

korgon (changes requested):
Not sure on the naming of the event, but I don't have firm feelings on it.

packages/snap-preact/src/Snap.tsx
dkonieczek: The handlers require the tracker via `this.tracker` in the file so we'd be passing things around. Didn't make this change

packages/snap-preact/src/setupEvents.ts
dkonieczek: Refactored setupEvents to clean things up

packages/snap-preact/src/setupEvents.ts
```diff
@@ -48,5 +34,59 @@ export const setupEvents = () => {
...
+const matchControllers = (
+	matchIds: (string | RegExp)[] | undefined,
+	conditionCallback?: (ctrl: AbstractController) => boolean
+): AbstractController[] => {
+	return Object.keys(window.searchspring.controller || {}).reduce((arr, id) => {
+		const controller = window.searchspring.controller[id] as AbstractController;
+
+		if (conditionCallback && !conditionCallback(controller)) {
+			return arr;
+		}
```
korgon, line 64: Maybe just use a filter on the return of `matchControllers` instead of callback?

korgon (changes requested):

korgon (approved):

PR #1160, opened 2024-09-10
refactor: client and store interface

dkonieczek:


korgon (approved):
We must be careful with regards to the breaking changes to the `beforeSearch`, `afterSearch` and `afterStore` middleware events.

PR #1169, opened 2024-09-24
checkbox a11y

chrisFrazier77:


korgon (changes requested):
Tests needs adjusting - we should assert that this new attribute exists or doesn't exist depending on the state.

korgon (changes requested):
Can we add a test assertion to verify that the checked state properly adjusts the aria-checked attribute?

PR #1181, opened 2024-10-23
add spanish translation

evanjohnso:


evanjohnso:
que pasoo?

korgon:
Lo sentimos mucho!

We have moved the template work to a fork, and will need this PR moved over there... 😬 
There are also additional components in the new fork that will require more translation. We can move this work over there for you in a new PR if you'd like.

https://github.com/searchspring/snap-1.0

korgon:
@chrisFrazier77 could you please move this work to the 1.0 fork and identify what additional props are missing since this translation was initially put forward?

PR #1196, opened 2024-11-12
bugFix: Tracking core result data 

chrisFrazier77:
Send core result data and not active display data in beacon tracking

### DIFF
--- a/packages/snap-controller/src/Recommendation/RecommendationController.test.ts
+++ b/packages/snap-controller/src/Recommendation/RecommendationController.test.ts
@@ -541,6 +541,10 @@ describe('Recommendation Controller', () => {
 		 */
 
 		const result = controller.store.results[0];
+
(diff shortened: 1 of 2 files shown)


packages/snap-controller/src/Recommendation/RecommendationController.test.ts
```diff
@@ -541,6 +541,10 @@ describe('Recommendation Controller', () => {
 		 */
 
 		const result = controller.store.results[0];
+
+		//this tests we send the core data and not the display data
+		result.mask.set({ id: 'custom' });
```
korgon, line 546: This test is fine, as It seems that we were always using `result.id` and not grabbing from `result.display.id`; but we may want to also test a change to the core mappings as well (as that has changed here).

korgon (changes requested):

korgon (approved):

PR #1208, opened 2024-11-18
Recommendation Back/Forward Cache

chrisFrazier77:


packages/snap-store-mobx/src/types.ts
```diff
@@ -145,6 +145,7 @@ export type RecommendationStoreConfig = StoreConfig & {
 	batchId?: number;
 	settings?: {
 		variants?: VariantConfig;
+		bfCacheRestore: boolean;
```
korgon, line 148: We have a similar setting for the `restorePosition` logic in the SearchController - it uses `restorePosition.onPageShow`. Perhaps we should use a similar name here? Maybe `searchOnPageShow` ?

korgon (changes requested):
Tests look to cover the controller creation use case. Have you tested the case when using the instantiator?

korgon (approved):

PR #1217, opened 2024-12-04
fix(urltranslator.ts): gracefully catch malformed uri errors

chrisFrazier77:


packages/snap-url-manager/src/Translators/Url/UrlTranslator.ts
```diff
@@ -148,13 +148,18 @@ export class UrlTranslator implements Translator {
...
 			.filter((v) => v)
 			.map((kvPair) => {
-				const [key, value] = kvPair.split('=').map((v) => decodeURIComponent(v.replace(/\+/g, ' ')));
-				return { key: key.split('.'), value, type: ParamLocationType.query };
+				try {
+					const [key, value] = kvPair.split('=').map((v) => decodeURIComponent(v.replace(/\+/g, ' ')));
+					return { key: key.split('.'), value, type: ParamLocationType.query };
+				} catch (err) {
+					console.error('Uncaught URIError: URI malformed', kvPair);
+					return { key: ['ss__delete'], value: 'ss__delete', type: ParamLocationType.query };
```
korgon, line 156: Could we just set the value to undefined instead of adding this additional case (since we remove empty params below) ?

packages/snap-url-manager/src/Translators/Url/UrlTranslator.ts
```diff
@@ -148,13 +148,18 @@ export class UrlTranslator implements Translator {
 			.split('&')
 			.filter((v) => v)
 			.map((kvPair) => {
-				const [key, value] = kvPair.split('=').map((v) => decodeURIComponent(v.replace(/\+/g, ' ')));
-				return { key: key.split('.'), value, type: ParamLocationType.query };
+				try {
+					const [key, value] = kvPair.split('=').map((v) => decodeURIComponent(v.replace(/\+/g, ' ')));
+					return { key: key.split('.'), value, type: ParamLocationType.query };
+				} catch (err) {
+					console.error('Uncaught URIError: URI malformed', kvPair);
```
korgon, line 155: Maybe a warning instead of error? Also some context around the error could be helpful:
```
console.warn('Snap UrlTranslator: URI malformed - ignoring parameter', kvPair);
```

packages/snap-url-manager/src/Translators/Url/UrlTranslator.ts
```diff
@@ -148,13 +148,18 @@ export class UrlTranslator implements Translator {
...
+					return { key: ['ss__delete'], value: 'ss__delete', type: ParamLocationType.query };
+				}
 			})
 			.filter((param) => {
 				// remove core fields that do not contain a value
```
korgon, line 160: Update to this comment might be nice for future us ;)

korgon (changes requested):
Some testing should be added around this. Also, is this a problem for hash parameters that we would need to also address?

packages/snap-url-manager/src/Translators/Url/UrlTranslator.test.ts
```diff
@@ -16,6 +16,29 @@ describe('UrlTranslator', () => {
 		expect(urlTranslator.serialize({})).toBe('/');
 	});
 
+	it('gracefully fails with malformed uri in hash', () => {
+		const url = 'http://example.com#/valid:true/utm_campaign:BFCM%2020%OFF%20Perfumes%20TEST%20(28.11.24)%20(clone)/valid2:truueee';
+		const urlTranslator = new UrlTranslator();
+		// @ts-ignore
+		expect(urlTranslator.parseHashString(url)).toStrictEqual([
+			{ key: ['http'], type: 'hash', value: '' },
+			{ key: ['example.com'], type: 'hash', value: '' },
```
korgon, line 25: I'm pretty sure that these should not be here... 🤔

korgon (approved):

PR #1230, opened 2024-12-18
Transform Badges Bugfix

chrisFrazier77:
bugfix in transformSearchResponse transforming badge fields

### DIFF
--- a/packages/snap-client/src/Client/transforms/searchResponse.ts
+++ b/packages/snap-client/src/Client/transforms/searchResponse.ts
@@ -214,7 +214,7 @@ transformSearchResponse.result = (rawResult: rawResult): SearchResponseModelResu
 	const attributes = Object.keys(rawResult)
 		.filter((k) => CORE_FIELDS.indexOf(k) == -1)
 		// remove 'badges' from attributes - but only if it is an object
-		.filter((k) => !(k == 'badges' && typeof rawResult[k] == 'object' && !Array.isArray(rawResult[k])))
+		.filter((k) => !(k == 'badges' && typeof rawResult[k] == 'object' && Array.isArray(rawResult[k]) && typeof rawResult[k]?.[0] == 'object'))
 		.reduce((attributes, key) => {
 			return {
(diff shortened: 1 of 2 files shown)


packages/snap-client/src/Client/transforms/searchResponse.ts
```diff
@@ -214,7 +214,7 @@ transformSearchResponse.result = (rawResult: rawResult): SearchResponseModelResu
 	const attributes = Object.keys(rawResult)
 		.filter((k) => CORE_FIELDS.indexOf(k) == -1)
 		// remove 'badges' from attributes - but only if it is an object
-		.filter((k) => !(k == 'badges' && typeof rawResult[k] == 'object' && !Array.isArray(rawResult[k])))
+		.filter((k) => !(k == 'badges' && typeof rawResult[k] == 'object' && Array.isArray(rawResult[k]) && typeof rawResult[k]?.[0] == 'object'))
```
korgon, line 217: Seems like the logic for checking `typeof rawResult[k] == 'object'` is unnecessary, as the next check is if it is an array.

packages/snap-client/src/Client/transforms/searchResponse.ts
```diff
@@ -242,7 +242,7 @@ transformSearchResponse.result = (rawResult: rawResult): SearchResponseModelResu
 			core: coreFieldValues,
 		},
 		attributes,
-		badges: typeof rawResult.badges == 'object' && !Array.isArray(rawResult.badges) ? rawResult.badges : [],
+		badges: typeof rawResult.badges == 'object' && Array.isArray(rawResult.badges) && typeof rawResult.badges[0] == 'object' ? rawResult.badges : [],
```
korgon, line 245: Seems like the logic for checking `typeof rawResult[k] == 'object'` is unnecessary, as the next check is if it is an array.

korgon (commented):
Nice test additions.

korgon (approved):

PR #1231, opened 2024-12-18
Infinite Store Refactor

korgon:
* changing the way products are concatenated when using the infinite feature


korgon:
Issue was found in code - changes incomming...

korgon:
Added logic to determine when to concatenate previous results; this logic is not bullet proof, but it is highly unlikely that the current and previous pagination data would ever be sequential unless paginating, so should be sufficient. The addition of having the previous search response in the store will also allow us to more easily add a "load previous" functionality in the future (by knowing pagination details.

dkonieczek (approved):

dkonieczek (approved):

PR #1241, opened 2025-01-17
fix(bundleselector.tsx): change checkbox size from number to px

chrisFrazier77:


korgon (changes requested):
While this will fix the default case, I think we should consider that others may also pass the checkbox Component just a number (which would cause problems).

Please see the originating issue for details on the actual fix.

packages/snap-preact-components/src/components/Molecules/Checkbox/Checkbox.tsx
```diff
@@ -13,8 +13,7 @@ import { Icon, IconProps } from '../../Atoms/Icon';
 import { useA11y } from '../../../hooks/useA11y';
 
 const CSS = {
-	checkbox: ({ size, color, theme }: Partial<CheckboxProps>) => {
-		const pixelSize = isNaN(Number(size)) ? size : `${size}px`;
+	checkbox: ({ pixelSize, color, theme }: Partial<CheckboxProps>) => {
```
korgon, line 16: `pixelSize` is not a CheckboxProp... which may cause issues when moving this to templates where these types are more strict... 🤔

packages/snap-preact-components/src/components/Molecules/Checkbox/Checkbox.tsx
```diff
@@ -99,7 +100,7 @@ export const Checkbox = observer((properties: CheckboxProps): JSX.Element => {
 		if (native) {
 			styling.css = [CSS.native(), style];
 		} else {
-			styling.css = [CSS.checkbox({ size, color, theme }), style];
+			styling.css = [CSS.checkbox({ pixelSize, color, theme }), style];
```
korgon, line 103: I think changing it to this would resolve the typing issue:
```
styling.css = [CSS.checkbox({ size: pixelSize, color, theme }), style];
```

korgon (commented):
See comments.

korgon (approved):

PR #1243, opened 2025-01-29
AC Bind Order

korgon:
* adjusting bind code so that it happens in proper order (sequentially)


korgon:
While the order of events is now correct, it interferes with TabManager usage, causing problems with the UrlManager subscribe function and searches being executed from that trigger. We cannot proceed with this change until all sites using TabManager are modified.

PR #1262, opened 2025-03-06
bugfix for infinite rerenders and researches when redirecting

chrisFrazier77:


korgon (changes requested):
I think we should probably add the same on line 103 (single result redirect).

korgon (commented):
Before we merge this in, I think we should consider that there is a lot of logic in existing implementations built around this store property.

For example, this property would typically indicate that we have executed a search and there is search data present, and so components would display that content.
https://github.com/searchspring/snapfu-scaffold-preact/blob/production/src/components/Content.jsx#L27

With this change is it possible that we will break some implementation displays or generate errors while the redirect is happening?

korgon (approved):
Tested several sites and my concerns don't seem to be valid.

PR #1263, opened 2025-03-06
OverlayBadge Styles

korgon:
* adding commonly used styles

### DIFF
--- a/packages/snap-preact-components/src/components/Molecules/OverlayBadge/OverlayBadge.tsx
+++ b/packages/snap-preact-components/src/components/Molecules/OverlayBadge/OverlayBadge.tsx
@@ -29,7 +29,6 @@ const CSS = {
 			position: 'relative',
 
 			'& .ss__overlay-badge__grid-wrapper': {
-				zIndex: 1,
 				pointerEvents: 'none',
 				display: 'grid',
 				position: 'absolute',
--- a/packages/snap-preact-components/src/components/Molecules/OverlayBadge/OverlayBadge.tsx
+++ b/packages/snap-preact-components/src/components/Molecules/OverlayBadge/OverlayBadge.tsx
@@ -29,6 +29,8 @@ const CSS = {
 			position: 'relative',
 
 			'& .ss__overlay-badge__grid-wrapper': {
+				zIndex: 1,
+				pointerEvents: 'none',
 				display: 'grid',
 				position: 'absolute',
 				top: 0,


dkonieczek (approved):

korgon:
@camspring - any other changes you were wanting to see here?

packages/snap-preact-components/src/components/Molecules/OverlayBadge/OverlayBadge.tsx
```diff
@@ -29,6 +29,8 @@ const CSS = {
 			position: 'relative',
 
 			'& .ss__overlay-badge__grid-wrapper': {
+				zIndex: 1,
```
chrisFrazier77, line 32: dont think we can do this z-index change, we have no idea what z-index sites that are currently using this feature have set on their other elements..

chrisFrazier77 (changes requested):
see comments

chrisFrazier77 (approved):

packages/snap-preact-components/src/components/Molecules/OverlayBadge/OverlayBadge.tsx
camspring: Good point @chrisFrazier77, I agree this is likely too opinionated.  @adriamurphy might be candidate for a toggle in the SCSS scaffolding as I've found it's been necessary for most of our implementations.  Thanks folks

adriamurphy: I talked this over with @korgon the other day and he showed me what Snap team did with cursor-pointer-events, which seems to work well. I agreed with him this was probably the better fix.

camspring: unfortunately the `z-index` part has already been merged and released as part of `0.64.0`: https://github.com/searchspring/snap/commit/fe8bcd6b6b8598e3fc13ea26c5f4f2d360e97117

so i guess it will need to be removed from `production` for a subsequent release

adriamurphy: @camspring What you're looking at appears to be an older commit. If you check this one, Kevin removed it:
https://github.com/searchspring/snap/commit/396bc6a93356f95853682853b4c5370663b77696

It is also not in the production / main branch which is currently on 0.64.0:
https://github.com/searchspring/snap/blob/main/packages/snap-preact-components/src/components/Molecules/OverlayBadge/OverlayBadge.tsx#L32

🤷‍♀️

camspring: @adriamurphy ah I see.  the reason I thought it was part of the `0.64.0` release is because there is a link to [commit fe8bcd6](https://github.com/searchspring/snap/commit/fe8bcd6b6b8598e3fc13ea26c5f4f2d360e97117) on the `0.64.0` release page: https://github.com/searchspring/snap/releases/tag/v0.64.0

commit `fe8bcd6` shows the `z-index` rule is still in there

image

PR #1281, opened 2025-04-01
test: add beacon 2.0 tracking e2e tests

dkonieczek:


packages/snap-controller/src/Search/SearchController.ts
```diff
@@ -337,15 +360,16 @@ export class SearchController extends AbstractController {
 					// TODO: in future, send as an interaction event
 				}
 			},
-			render: (result: Product | Banner) => {
-				if (this.events.product[result.id]?.render) {
+			render: (result?: Product) => {
```
korgon, line 363: I think `result` should not be optional here.

Maybe just need to undo the changes to this method - since we are not using this for no results conditions any longer.

packages/snap-controller/src/Autocomplete/AutocompleteController.ts
```diff
@@ -54,7 +56,7 @@ type AutocompleteTrackMethods = {
 	product: {
 		clickThrough: (e: MouseEvent, result: Product) => void;
 		click: (e: MouseEvent, result: Product | Banner) => void;
-		render: (result: Product) => void;
+		render: (result?: Product) => void;
```
korgon, line 59: Should not be optional any longer.

packages/snap-controller/src/Autocomplete/AutocompleteController.ts
```diff
@@ -106,6 +108,21 @@ export class AutocompleteController extends AbstractController {
...
+					this.track.product.render();
+				}
+				controller.store.results.forEach((result) => {
+					if (result.type === 'product') {
+						this.track.product.render(result as Product);
+					}
+				});
+			}
+		});
+
```
korgon, line 125: Need to change to using same code as used in the SearchController.

packages/snap-controller/src/Autocomplete/AutocompleteController.ts
```diff
@@ -165,13 +182,14 @@ export class AutocompleteController extends AbstractController {
...
+				const key = result?.id || 'noresults';
+				if (this.events.product[key]?.render) return;
 
-				const data = getAutocompleteSchemaData({ params: this.params, store: this.store, results: [result] });
+				const data = getAutocompleteSchemaData({ params: this.params, store: this.store, results: result ? [result] : [] });
 				this.tracker.events.autocomplete.render({ data, siteId: this.config.globals?.siteId });
-				this.events.product[result.id] = this.events.product[result.id] || {};
-				this.events.product[result.id].render = true;
+				this.events.product[key] = this.events.product[key] || {};
+				this.events.product[key].render = true;
```
korgon, line 192: Should revert, no longer using this for no results conditions.

packages/snap-controller/src/Recommendation/RecommendationController.ts
```diff
@@ -5,15 +5,16 @@ import { AbstractController } from '../Abstract/AbstractController';
...
+import type { Item, Product as BeaconProduct, RecommendationsAddtocartSchemaData, RecommendationsSchemaData } from '@searchspring/beacon';
+import type { Next } from '@searchspring/snap-event-manager';
 
 // TODO: change return types to void
 type RecommendationTrackMethods = {
 	product: {
 		clickThrough: (e: MouseEvent, result: Product) => void;
 		click: (e: MouseEvent, result: Product) => void;
-		render: (result: Product) => void;
+		render: (result?: Product) => void;
```
korgon, line 17: Should not be optional.

packages/snap-controller/src/Recommendation/RecommendationController.ts
```diff
@@ -73,6 +74,21 @@ export class RecommendationController extends AbstractController {
...
+					this.track.product.render();
+				}
+				controller.store.results.forEach((result) => {
+					if (result.type === 'product') {
+						this.track.product.render(result as Product);
+					}
+				});
+			}
+		});
+
```
korgon, line 91: Should use similar technique as was refactored in SearchController - build out event from store data and handle no results case.

packages/snap-controller/src/Recommendation/RecommendationController.ts
```diff
@@ -121,18 +137,19 @@ export class RecommendationController extends AbstractController {
...
+				const key = result?.id || 'noresults';
+				if (this.events.product[key]?.render) return;
 
-				const data = getRecommendationsSchemaData({ store: this.store, results: [result] });
+				const data = getRecommendationsSchemaData({ store: this.store, results: result ? [result] : [] });
 				this.tracker.events.recommendations.render({ data, siteId: this.config.globals?.siteId });
-				this.events.product[result.id] = this.events.product[result.id] || {};
-				this.events.product[result.id].render = true;
+				this.events.product[key] = this.events.product[key] || {};
+				this.events.product[key].render = true;
```
korgon, line 147: Should probably revert, no results condition should not be handled via this method.

packages/snap-controller/src/Autocomplete/AutocompleteController.ts
```diff
@@ -867,12 +909,14 @@ function getAutocompleteSchemaData({
...
-			},
-		],
+		sort: store.sorting.options
+			.filter((sort) => sort.active)
+			?.map((sort) => {
+				return {
+					field: sort.field,
+					dir: sort.direction as AutocompleteSchemaDataSortInnerDirEnum,
+				};
+			}),
```
korgon, line 919: Unfortunately, after looking into this further, it appears the active property is set on the sort even if it is not "applied". We will likely need to use the request parameters instead to find the "applied" sorts.

image

packages/snap-controller/src/Search/SearchController.ts
```diff
@@ -744,12 +785,14 @@ function getSearchSchemaData({
...
-			},
-		],
+		sort: store.sorting.options
+			.filter((sort) => sort.active)
+			?.map((sort) => {
+				return {
+					field: sort.field,
+					dir: sort.direction as AutocompleteSchemaDataSortInnerDirEnum,
+				};
+			}),
```
korgon, line 795: As mentioned in AutocompleteController comment, we will likely need to use the request parameters instead.

packages/snap-preact-demo/tests/cypress/e2e/tracking/track.cy.js
```diff
@@ -1,382 +1,237 @@
...
+			expect(context).to.have.property('timestamp').to.be.a('string');
+			expect(context).to.have.property('userAgent').to.be.a('string');
+			expect(context).to.have.property('userId').to.be.a('string');
+
+			expect(response.body).to.have.property('success').to.equal(true);
+		});
 	});
-	it('tracked shopper login', () => {
+
+	it('tracked search render, impression, clickthrough', () => {
```
korgon, line 30: We might want to assert that they all have the same context data?

packages/snap-preact-demo/tests/cypress/e2e/tracking/track.cy.js
```diff
@@ -1,382 +1,237 @@
...
+		cy.wait(`@beacon2/search/render`).then(({ request, response }) => {
+			expect(response.body).to.have.property('success').to.equal(true);
+
+			const { context, data } = JSON.parse(request.body);
+			expect(context).to.be.an('object');
+			expect(data).to.have.property('results').to.be.an('array').length.greaterThan(0);
+			expect(data.results[0]).to.have.property('uid').to.be.a('string');
+			expect(data).to.have.property('merchandising').to.be.an('object');
+			expect(data).to.have.property('pagination').to.be.an('object');
+			expect(data).to.have.property('sort').to.be.an('array').length.greaterThan(0);
```
korgon, line 42: This should be an array equal to zero - or undefined, as no sort is applied on base landing page.

packages/snap-preact-demo/tests/cypress/e2e/tracking/track.cy.js
```diff
@@ -1,382 +1,237 @@
...
 		cy.visit('https://localhost:2222');
 
-		cy.waitForBundle().then((searchspring) => {
-			expect(searchspring).to.exist;
+		cy.wait(`@beacon2/search/render`).then(({ request, response }) => {
+			expect(response.body).to.have.property('success').to.equal(true);
+
+			const { context, data } = JSON.parse(request.body);
+			expect(context).to.be.an('object');
+			expect(data).to.have.property('results').to.be.an('array').length.greaterThan(0);
```
korgon, line 38: To improve the test we could assert the result data agrees fully with the result store data.

packages/snap-preact-demo/tests/cypress/e2e/tracking/track.cy.js
```diff
@@ -1,382 +1,237 @@
...
-				expect(beacon.context).to.be.an('object').include.key('shopperId');
-				expect(beacon.context.shopperId).to.equal(shopperId);
-			});
+		cy.scrollTo(0, 500); // Scroll down to trigger impressions
+
+		cy.wait(`@beacon2/search/impression`, { timeout: 10000 }).then(({ request, response }) => {
+			expect(response.body).to.have.property('success').to.equal(true);
+
+			const { data } = JSON.parse(request.body);
+			expect(data).to.have.property('results').to.be.an('array').length.greaterThan(0);
```
korgon, line 52: We should be explicit instead of just greater than zero - exactly how many results do we expect to be here?

packages/snap-preact-demo/tests/cypress/e2e/tracking/track.cy.js
```diff
@@ -1,382 +1,237 @@
...
+		});
+
+		// Click on first product
+		cy.get('.ss__result a').first().click({ force: true });
+
+		cy.wait(`@beacon2/search/clickthrough`).then(({ request, response }) => {
+			expect(response.body).to.have.property('success').to.equal(true);
+
+			const { data } = JSON.parse(request.body);
+			expect(data).to.have.property('results').to.be.an('array').length(1);
```
korgon, line 62: Should we assert the correct product data was sent?

packages/snap-preact-demo/tests/cypress/e2e/tracking/track.cy.js
korgon: In addition, I think we should assert the data object is _exactly_ or strictly equal to another object - so that we are checking that we are not sending properties that we shouldn't be. These assertions over expected properties are good, but wouldn't catch additional incorrect properties.

packages/snap-preact-demo/tests/cypress/e2e/tracking/track.cy.js
korgon, line 1: I've made a few comments on the search event test - but many of the callouts there should be considered for the other tests cases.

Most importantly I think we should be asserting that specific data is being sent. And any assertions should be checking the entire object, not just for certain properties on it, so that we can catch if any unnecessary or unexpected properties exist when they shouldn't.

korgon (changes requested):
Initial review - Github is showing me a "Refresh" button, so there must be additional commits to look at... I'll have to check that in a follow up review.

korgon:
An edge case that you should test for and will need to likely address is when using infinite scroll - currently the middleware method of sending the render I believe would be problematic, as we would be sending previous page result data in render events for subsequent pages - a mechanism will need to be found to prevent this.

PR #1294, opened 2025-04-16
feat: add withTracking and useTracking, post beacon launch fixes

dkonieczek:


packages/snap-controller/src/Search/SearchController.test.ts
```diff
@@ -433,10 +433,10 @@ describe('Search Controller', () => {
 		expect(searchfn).toHaveBeenCalledTimes(1);
 
 		await controller.search();
-		expect(searchfn).toHaveBeenCalledTimes(2); // a 2nd search is called due to redirectResponse setting (store.loaded)
+		expect(searchfn).toHaveBeenCalledTimes(1); // a 2nd search is called due to redirectResponse setting (store.loaded)
```
korgon, line 436: The comment here is no longer correct.

packages/snap-preact-components/src/components/Molecules/Result/Result.test.tsx
korgon, line 1: 🤔 ... was the mockIntersectionObserver needed for this test to pass - doesn't seem like we do any assertions on impressions...

packages/snap-preact-components/src/components/Trackers/ResultTracker/ResultTracker.tsx
korgon, line 1: To keep thinks consistent, should this component utilize the new tracking HOC?

packages/snap-preact-components/src/providers/withTracking.tsx
```diff
@@ -0,0 +1,61 @@
...
+			}
+		}
+
+		const currentRef = trackingRef.current;
+		if (currentRef) {
+			const handleClick = (e: MouseEvent) => {
+				controller?.track.product.click(e, result as Product);
+			};
+			currentRef.setAttribute('sstracking', 'true');
+			currentRef.removeEventListener('click', handleClick);
```
korgon, line 47: I'm not sure that this remove event listener call will work if we are trying to remove a previously set handler. This is because the reference to the `handleClick` function will have been re-created in the above lines, and the reference to the original `handleClick` function that was previously attached will have been lost.

packages/snap-preact-components/src/hooks/useTracking.tsx
korgon, line 1: This ties back to the ResultTracker comment above - I'm not sure it is possible, but could `ResultTracker`, `useTracking` and `withTracking` utilize the same code instead of repeating it?

korgon (changes requested):

packages/snap-preact-components/src/components/Molecules/Result/Result.test.tsx
dkonieczek: The Result component now uses withTracking and that's using IntersectionObserver so this was failing.

packages/snap-preact-components/src/components/Trackers/ResultTracker/ResultTracker.tsx
dkonieczek: I thought this would be kept as an alternative to the HOC

packages/snap-preact-components/src/providers/withTracking.tsx
korgon, line 1: I still think we should keep the removeEventListener - however, the way it was setup would not have actually removed it. I think in order for it to work we would need to lift the handler up out of the `withTracking` function so that a reference to the function is maintained. This would require some refactoring though since the handler needs a reference to the controller.

packages/snap-preact-components/src/components/Trackers/ResultTracker/ResultTracker.tsx
korgon: I think I am ok with it as an alternative, but can it use the HOC under the hood (or the hook) so that all of this logic only exists in one place? (similar to my last comment in this review)

korgon:
Test is failing... perhaps now is a good time to re-attempt making this test stable.

dkonieczek:
The past 4 actions have been successful. 🤞 

removeEventListener requires a reference to a function but then you can't pass in a controller and I don't see a way to refactor this. I think it would be fine without removing or keeping it as is. A new instance of withTracking is created for each Result element so should be fine? Or maybe upon rerender, the instance of withTracking isn't recreated and having the removeEventListener call would then make sense? 

I also refactored some of the same useIntersectionAdvanced code into a single createImpressionObserver function. At least the observer parameters are defined in a single spot now. The usage across the 3 varies so that why this also doesn't include the tracking call itself.

korgon (dismissed):
In a recent discussion we talked about the possibility of removing options for doing the tracking integration (eg. getting rid of the hook or other usages). However, we need to keep `<ResultTracker>` for legacy purposes (for profile and result recommendations tracking as many styles exist associated with these wrappers).

That leaves the HOC and the hook - and I think here we should keep both.

packages/snap-store-mobx/src/Autocomplete/Stores/AutocompleteQueryStore.ts
```diff
@@ -1,12 +1,17 @@
...
+	AutocompleteResponseModelAllOfAutocomplete,
+	SearchResponseModelSearchMatchTypeEnum,
+} from '@searchspring/snapi-types';
 import type { AutocompleteStoreConfig, StoreServices } from '../../types';
 
 export class AutocompleteQueryStore {
 	public query?: Query;
 	public originalQuery?: Query;
 	public correctedQuery?: Query;
+	public matchType?: SearchResponseModelSearchMatchTypeEnum;
```
korgon, line 14: Seems we may need to alter SnapiOAS so that it includes `primary` as a possible type?


https://github.com/searchspring/snapi-types/blob/v0.1.38/models/SearchResponseModelSearch.d.ts#L48-L50

packages/snap-tracker/src/Tracker.ts
```diff
@@ -87,7 +86,8 @@ export class Tracker extends Beacon {
 							this.track.product.view(item, siteId);
 							break;
 						case 'searchspring/track/cart/view':
-							this.track.cart.view({ items }, siteId);
+							console.warn('searchspring/track/cart/view attribute tracking is deprecated and will be removed');
```
korgon, line 89: Is it technically removed at this point?

korgon (changes requested):
Need to see if testing is more reliable.

packages/snap-preact-components/src/providers/withTracking.tsx
korgon: This is one advantage the hook method has over the HOC - it can correctly clean up after itself.

PR #1311, opened 2025-05-14
feat: add cart infering, update tracking documentation

dkonieczek:


packages/snap-tracker/src/Tracker.ts
```diff
@@ -174,6 +174,110 @@ export class Tracker extends Beacon {
...
+							cart: [],
+						},
+					});
+				}
+				this.storage.cart.clear();
+			} else if (cart.length) {
+				// length check here to be able to sent error event if invalid array of objects is provided
+				const currentCart: Product[] = cart
+					.filter((item) => typeof item === 'object' && (item.uid || item.sku || item.childUid || item.childSku) && item.qty && item.price)
+					.map((item): Product => {
```
korgon, line 196: I'm not sure I understand the map usage here - seems to map to same properties as `item`... 🤔

korgon (changes requested):

korgon (dismissed):

korgon (dismissed):
Looks good.

packages/snap-controller/src/Search/SearchController.ts
```diff
@@ -447,6 +454,29 @@ export class SearchController extends AbstractController {
 		return params;
 	}
 
+	createResultSchemaMapping = (request: SearchRequestModel, response: [MetaResponseModel, SearchResponseModel]) => {
```
korgon, line 457: I think I'd prefer this not be a class method if possible - as I can see it going away at some point.

packages/snap-controller/src/Search/SearchController.ts
```diff
@@ -759,7 +788,17 @@ function getSearchAddtocartSchemaData({
...
 
-function getSearchSchemaData({ params, store, results }: { params: SearchRequestModel; store: SearchStore; results?: Product[] }): SearchSchemaData {
+function getSearchSchemaData({
+	params,
+	store,
+	results,
+	response,
+}: {
+	params: SearchRequestModel;
+	store: SearchStore;
```
korgon, line 798: Does it make sense to use the store values as fallback values here? If not, it seems we shouldn't really need the store.

packages/snap-controller/src/Search/SearchController.ts
```diff
@@ -89,6 +89,7 @@ export class SearchController extends AbstractController {
 			}
 		>;
 	} = { product: {} };
+	private schemaMap: Record<string, SearchSchemaData> = {};
```
korgon, line 92: If possible it would be good not to expose this as a property on the class (private in TS is not enough to prevent people from using it or modifying its contents).

korgon (commented):
Let's discuss these comments.

korgon (approved):

PR #1335, opened 2025-07-03
feat(searchresultstore): adding support for api variants

chrisFrazier77:


korgon (changes requested):
Ran into an error when testing on a site (z7h1jh) where the `ss_variants` field has been removed from display in recs, but the variants core mapping was set.

The initial error is resolved by merging in develop, the next one is resolved with the following change:

Image

I was going to commit said changes (merge develop and line above) - but that makes for weird PR review; so I'll leave it to you.

PR #1355, opened 2025-07-31
docs(docs): reskin docs to new athos branding

chrisFrazier77:


assets/styles.css
```diff
@@ -17,15 +17,15 @@ body {
 }
 
 #search-container {
-	color: hsl(247deg 67% 33%);
+	color: #1D4990;
```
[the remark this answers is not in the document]
korgon: @chrisFrazier77 this is a good suggestion that we should consider.

images/flat-minimal-logo-snap.svg
korgon, line 1: The "flat-minimal-logo-snap" logo is meant to be flat - no shadow, no white outline or border.

images/logo-snap.svg
korgon, line 1: It is probably fine, but we may want a little more margin on the bottom for this image.

images/flat-logo-snap.svg
korgon, line 1: This image should have the shadow removed - it looks like the normal logo as is.

PR #1383, opened 2025-10-16
Brand Categories

korgon:
* adding 'brand' and 'manufacturer' to the list of known background filters


packages/snap-controller/src/Search/SearchController.ts
dkonieczek: I think this block needs to not happen for this case since the value would be a string and not any of the `BACKGROUND_FILTERS_VALUE_FLAGS` values?

PR #1384, opened 2025-10-16
Release 0.69.0

### LINKED ISSUE #1367: Add import lines in Storybook
Something like this for each component readme.
```
import { Pagination } from '@searchspring/snap-preact-components';
```

### LINKED ISSUE #1372: Add hierarchy filters to filter summary
Currently hierarchy filters are not added to the `controller.store.filters array` so they don't show up in the filter summary. 

It isn't very intuitive to add every other type of filter to the summary but omit hierarchy filters. Clients ask for this often enough that I now typically add the following middleware in an `afterStore` callback by default in my projects:

```javascript
// add hierarchy filter to filter summary
if (controller.store.facets.length > 0) {
	const dataDelimiter = '>';  // choose delimiter in data
	const displayDelimiter = ' / ';  // choose delimiter for label
	const showFullPath = false;  // display full hierarchy path or just the current level 

	controller.store.facets.forEach((facet) => {
		if (facet.display === 'hierarchy' && facet.filtered && facet.values?.length > 0) {
			// check for filtered value to make sure it's a user-applied filter and not a background filter
			const filteredValue = facet.values.find((v) => v.filtered);
			if (filteredValue) {
				const crumb = controller.urlManager.urlState.filter[facet.field];
				if (crumb) {
					const categoryParts = crumb[0].split(dataDelimiter);
					const childCategory = categoryParts[categoryParts.length - 1];
					const hierarchyFilter = {
						facet: {
							field: facet.field,
							label: facet.label,
						},
						label: `${facet.label}: ${showFullPath ? crumb[0].replaceAll(dataDelimiter, displayDelimiter) : childCategory}`,
						value: {
							label: showFullPath ? crumb[0].replaceAll(dataDelimiter, displayDelimiter) : childCategory,
							value: crumb[0],
						},
						url: {
							link: controller.urlManager.remove('filter.' + facet.field).link,
						},
					};
					controller.store.filters.push(hierarchyFilter);
				}
			}
		}
	});
}
```

I think this should be enabled by default and have the following options and default values:

- Data delimiter: `">"`
- Display delimiter for labels: `" / "`
- Boolean to display the full hierarchy path: `false` (display currently level only by default since hierarchies tend to product very long labels).

### LINKED ISSUE #1374: Safari "Block All Cookies" setting
When the "Block All Cookies" setting is set in safari, Snap breaks down and no longer works. This should work just fine and treat every page as a brand new customer with no cookies or stored history.

### LINKED ISSUE #1380: Custom UID
Need ability to specify a specific UID field to use for the core UID to send to beacon. This is necessary due to explosion of products at the variant level. Future API will have `parentId` as a core field. Until then the configuration will be necessary.

korgon:
closes #1367
closes #1372
closes #1374
closes #1380

### DIFF
--- a/packages/snap-controller/src/Search/README.md
+++ b/packages/snap-controller/src/Search/README.md
@@ -15,7 +15,7 @@ The `SearchController` is used when making queries to the API `search` endpoint.
 | settings.facets.trim | facets that do not change results will be removed | true |   |
 | settings.facets.autoOpenActive | setting for "auto open" functionality for facets that are filtered (active), collapsed, and have no stored data | true |   |
 | settings.facets.fields | object keyed by individual facet fields for configuration of any settings.facets options | ➖ |   |
-| settings.filters.hierarchy.showInSummary | boolean to enable/disable selected hierarchy facets from showing in the filters  | false |   |
+| settings.filters.hierarchy.enabled | boolean to enable/disable selected hierarchy facets from showing in the filters  | false |   |
 | settings.filters.hierarchy.showFullPath | boolean to show the full hierarchy path in the filter  | false |   |
 | settings.filters.hierarchy.displayDelimiter | string to adjust the delimiter between each level of the full hierarchy path | ' / ' |   |
 | settings.history.max | how many search terms should be kept in the history store | 25 |   | 


packages/snap-controller/src/Search/README.md
```diff
@@ -15,6 +15,9 @@ The `SearchController` is used when making queries to the API `search` endpoint.
 | settings.facets.trim | facets that do not change results will be removed | true |   |
 | settings.facets.autoOpenActive | setting for "auto open" functionality for facets that are filtered (active), collapsed, and have no stored data | true |   |
 | settings.facets.fields | object keyed by individual facet fields for configuration of any settings.facets options | ➖ |   |
+| settings.filters.hierarchy.showInSummary | boolean to enable/disable selected hierarchy facets from showing in the filters  | false |   |
```
chrisFrazier77: wrong...

chrisFrazier77 (approved):

PR #1401, opened 2025-11-20
FacetSlider Enhancements

korgon:
* adding 'separateHandles' prop
* adjusting onChange prop to allow for mutation of values as needed


packages/snap-preact-components/src/components/Molecules/FacetSlider/FacetSlider.stories.tsx
```diff
@@ -135,6 +135,16 @@ export default {
 			},
 			control: { type: 'boolean' },
 		},
+		separateHandles: {
+			description: 'separates slider handles by one step value to prevent min and max from being equal',
```
chrisFrazier77: seems to work. should we add a unit test for the new prop?

PR #1413, opened 2025-12-09
feat(snap.tsx): adding support for configurable initiator

chrisFrazier77:


korgon (changes requested):
Looks good, but we should probably update the demo to include the new `modern.ts` entry point for the modern build eh?

korgon (approved):

PR #1426, opened 2026-01-13
fix(searchresultstore): adding support for option element to be used …

chrisFrazier77:
…in realtime variants

### DIFF
--- a/packages/snap-store-mobx/src/Search/Stores/SearchResultStore.ts
+++ b/packages/snap-store-mobx/src/Search/Stores/SearchResultStore.ts
@@ -85,9 +85,7 @@ export class SearchResultStore extends Array<Product | Banner> {
 					const attr = elem.getAttribute(VARIANT_ATTRIBUTE);
 					if (attr) {
 						const [option, value] = attr.split(':');
-						if (!option || !value) {
-							console.error('Error!: realtime variant is missing option or value (option:value)!', elem, attr);
-						} else if (option && value) {
+						if (option && value) {
 							options[option.toLowerCase()] = [value.toLowerCase()];
 						}
 					}
@@ -764,9 +762,12 @@ function variantOptionClick(elem: Element, variantConfig: VariantConfig, results
 	const attr = elem.getAttribute(VARIANT_ATTRIBUTE);
 	if (attr) {
 		const [option, value] = attr.split(':');
-		options[option.toLowerCase()] = [value.toLowerCase()];
-
-		makeVariantSelections(variantConfig, options, results);
+		if (!option || !value) {
+			console.error('Error!: realtime variant is missing option or value (option:value)!', elem, attr);
+		} else {
+			options[option.toLowerCase()] = [value.toLowerCase()];
+			makeVariantSelections(variantConfig, options, results);
+		}
 	}
 }
 


packages/snap-store-mobx/src/Search/Stores/SearchResultStore.ts
```diff
@@ -45,14 +45,30 @@ export class SearchResultStore extends Array<Product | Banner> {
...
 		if (variantConfig?.realtime?.enabled) {
 			// attach click events - ONLY happens once (known limitation)
 			if (!loaded && results.length) {
 				document.querySelectorAll(`[${VARIANT_ATTRIBUTE}]`).forEach((elem) => {
 					if (variantConfig?.field && !variantConfig?.realtime?.enabled === false) {
-						elem.addEventListener('click', () => variantOptionClick(elem, variantConfig, results));
+						if (elem.tagName == 'OPTION') {
+							const parentSelect = elem.closest('select');
+							if (parentSelect) {
+								parentSelect.addEventListener('change', (e) => {
```
korgon: We need to prevent each option from adding a `change` event listener on the select.

chrisFrazier77: good catch

packages/snap-store-mobx/src/Search/Stores/SearchResultStore.ts
```diff
@@ -45,35 +45,53 @@ export class SearchResultStore extends Array<Product | Banner> {
...
-					if (variantConfig?.field && !variantConfig?.realtime?.enabled === false) {
-						elem.addEventListener('click', () => variantOptionClick(elem, variantConfig, results));
+					if (elem.tagName == 'OPTION') {
+						const parentSelect = elem.closest('select');
+						if (parentSelect) {
+							if (!processedSelects.has(parentSelect)) {
+								processedSelects.add(parentSelect);
+								parentSelect.addEventListener('change', (e) => {
+									const val = (e.target as HTMLSelectElement)?.value;
+									const clickedOption = Array.from(document.querySelectorAll(`[${VARIANT_ATTRIBUTE}]`)).filter(
```
korgon: This could be improved by only doing a querySelectorAll from within the `parentSelect` element instead of in the entire document.

packages/snap-store-mobx/src/Search/Stores/SearchResultStore.ts
```diff
@@ -45,35 +45,53 @@ export class SearchResultStore extends Array<Product | Banner> {
...
+				const processedSelects = new Set<Element>();
 				document.querySelectorAll(`[${VARIANT_ATTRIBUTE}]`).forEach((elem) => {
-					if (variantConfig?.field && !variantConfig?.realtime?.enabled === false) {
-						elem.addEventListener('click', () => variantOptionClick(elem, variantConfig, results));
+					if (elem.tagName == 'OPTION') {
+						const parentSelect = elem.closest('select');
+						if (parentSelect) {
+							if (!processedSelects.has(parentSelect)) {
+								processedSelects.add(parentSelect);
+								parentSelect.addEventListener('change', (e) => {
```
korgon: Enhancement: When we attach this handler should we also check for a `selected` option and auto-add the "ss-variant-option-selected" attribute? This could make the setup here a bit simpler, as currently pre-selection will not work without that attribute.

PR #1434, opened 2026-01-22
Release v0.73.0

dkonieczek:
* Auto-beaconing using latest beacon version and spec
* Adds merchandising banner impressions and click events
* Adds inline banner impressions and click events
* Adds category.json endpoint
* Adds 'input' and 'source' parameters to autocomplete api


korgon (approved):
👋🏻

PR #1444, opened 2026-02-12
Ready controller search

chrisFrazier77:


packages/snap-preact/src/Instantiators/RecommendationInstantiator.tsx
```diff
@@ -340,8 +339,8 @@ async function readyTheController(
 		instance.middleware.forEach((middleware) => controller.on(middleware.event, ...middleware.func));
 	}
 
-	// run a search on the controller if it has not yet and it is not currently
-	if (!controller.store.loaded && !controller.store.loading) {
```
camspring: @chrisFrazier77  I've noticed this double condition in the `Bundle.jsx` scaffolding that is generated when you run `snapfu init recs`.  I never quite understood why were were checking both `loaded` and `loading`.  In the context of a recs component, eg: `Bundle.jsx` are both still required?

korgon: The purpose is to prevent running another search when it is already loaded - but really it is just `loading` that matters the most, as we wouldn't need to run a search if one is already loading.

PR #1446, opened 2026-02-18
With tracking use callback

dkonieczek:


packages/snap-controller/src/Autocomplete/AutocompleteController.ts
korgon: Is this still needed? And if so, should we just use `responseId`?

PR #1473, opened 2026-04-02
fix(searchresultstore.ts): added a new sort for the searchresultstore…

### LINKED ISSUE #1453: Incorrect variant order in `result.variants.selections[].values` on Bundle seed product after realtime update
When the Bundle seed product selection is updated by realtime variants functionality, the list of available options in `result.variants.selections[].values` can sometimes appearing "out of order".  Seems to only happen when there are multiple `ss-variant-option-selected` and `ss-variant-option` attributes on the main product, eg: "color" and "size". 

For example, let's say the default **color** variant "BLUE" has the following available **size** options:
```
M, L, XL, XXL
```

The `ss-variant-option-selected` attribute is targeting variant "color:RED", which has the following available **size** options:
```
XS, S, M, L, XL, XXL
```

When realtime variants functionality selects the "RED" **color** variant, its additional **size** options will be added to the end of `result.variants.selections[].values` and appear out of order:
```
M, L, XL, XXL, XS, S
```

Relevant Slack thread: https://athoscommerce.slack.com/archives/CQ3F8V2E5/p1770963853835209

chrisFrazier77:
… that sorts values correctly

### DIFF
--- a/packages/snap-store-mobx/src/Search/Stores/SearchResultStore.test.ts
+++ b/packages/snap-store-mobx/src/Search/Stores/SearchResultStore.test.ts
@@ -1615,6 +1615,78 @@ describe('SearchResultStore', () => {
 				const previouslySelectedSelections = variants.selections.map((selection) => selection.selected?.value);
 				expect(previouslySelectedSelections).toStrictEqual(['Desert', '30', '34']);
 			});
+
+			it('sorts selection values correctly when first color group has incomplete sizes', () => {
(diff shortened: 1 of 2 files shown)


korgon (commented):
This solution seems a bit fragile in that we are assuming that the variant with the most options will account for all options - but any options not within that variant would still be added to the end right?

I suppose the only way to preserve a fixed order in those cases would be some manual sorting on the front end?

